PR #47, opened 2016-08-15
Enable algo to configure a local server without having to use an ssh connection

computerality:
and update documentations with minimal installation requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,15 @@ Start the deploy and follow the instructions:
 
 When the process is done, you can find `.mobileconfig` files and certificates in the `configs` directory. Send the `.mobileconfig` profile to users with Apple devices. Note that profile installation is supported over AirDrop. Do not send the mobileconfig file over plaintext (e.g., e-mail) since it contains the keys to access the VPN. For those using other clients, like Windows or Android, securely send them the X.509 certificates for the server and their user.
 
+
+### Initial Deployment on Local Servers
+When installing algo on an existing system such as Ubuntu 16.04, you only need to install ansible using the commands below:
+
+ ```
+ sudo apt-get install software-properties-common && sudo apt-add-repository ppa:ansible/ansible
+ sudo apt-get update && sudo apt-get install ansible
+ ```
+
 ### User Management
 
 If you want to add or delete users, update the `users` list in `config.cfg` and run the command: 
--- a/algo
+++ b/algo
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 p12_export_password () {
   echo -n "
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -62,6 +62,15 @@ Start the deploy and follow the instructions:
 
 When the process is done, you can find `.mobileconfig` files and certificates in the `configs` directory. Send the `.mobileconfig` profile to users with Apple devices. Note that profile installation is supported over AirDrop. Do not send the mobileconfig file over plaintext (e.g., e-mail) since it contains the keys to access the VPN. For those using other clients, like Windows or Android, securely send them the X.509 certificates for the server and their user.
 
+
+### Initial Deployment on Local Servers
+When installing algo on an existing system such as Ubuntu 16.04, you only need to install ansible using the commands below:
+
+ ```
+ sudo apt-get install software-properties-common && sudo apt-add-repository ppa:ansible/ansible
+ sudo apt-get update && sudo apt-get install ansible
```
dguido: Installing ansible with pip is generally the best strategy here. What is software-properties-common for?

algo
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
dguido: :-(

algo
```diff
@@ -31,12 +31,29 @@ Enter the number of your desired provider
...
 }
 
 user_management () {
   p12_export_password
-  ansible-playbook users.yml -e easyrsa_p12_export_password=${P}
+  # Use a local connection for non-cloud installations
+  echo -n "
+  What provider would you like to use?
+    1. Remote Server
+    2. Local Server
```
dguido: Oh, I see. Maybe we can simply add an option to the initial prompt rather than a second one here.

README.md
computerality: When installing ansible from pip, it depends on python's crytography library which requires  requires these packages being installed: build-essential, python-dev, libssl-dev. The last one there is openssl. Installing from apt is a little easier.

algo
computerality: On Ubuntu, /bin/sh is dash instead of bash. And that uses a different version of the read command. You get this message: "./algo: 7: read: Illegal option -s" on Ubuntu 16.04

computerality: A better way to fix this might be switching to prompting for the user cert password within ansible with the `private: yes` option set instead of reading it from the shell directly.

algo
gunph1ld: We need to decrease usage `case`, `if` and other operators in the initial script

algo
gunph1ld: Then we need to use `#!/usr/bin/env bash`

gunph1ld:
There are some remarks:
1. Clean up the `algo` script and move additional options to ansible (we can use localhost as the inventory_hostname) 
2. Change bash path

README.md
dguido: Ah ok, I get this now. Installing software-properties-common will install the apt-add-repository command, and the ppa:ansible/ansible is required because Ubuntu 16.04 only comes with Ansible 2.0.0.2.

algo
dguido: I think I would prefer adding this prompt to Ansible and removing the shell commands from our script that dash does not support. Then we can keep it at /bin/sh.

PR #55, opened 2016-08-17
Digitalocean: Add new region blr1 Bangalore

vishnunaini:
Digitalocean has added a new region few months ago and this pull changes the config file for digitalocean

See https://www.digitalocean.com/company/blog/introducing-our-bangalore-region-blr1/


dguido:
:+1:

PR #69, opened 2016-08-20
Add HTTPS support in adblock.sh

colinmahns:
Just a quick change to adblock.sh. First commit changes http://www.mvps.org/winhelp2002/hosts.txt to point to http://winhelp2002.mvps.org/hosts.txt, there's a server-side redirect to the latter page when accessing the former page.

I've also added HTTPS to hosts-file.net, malwaredomainlist.com, and adaway.org as each of these domains support it with a valid certificate. 

I've tested this on my instance of algo to ensure there is no breakage, but feel free to doublecheck my PR :)


dguido:
:+1:

PR #125, opened 2016-11-06
Add missing playbook tags

cernekee:
I believe this broke in d052cb8e772af160c8a13b934f7146d53d1a7876

If the common playbook is invoked with the "cloud" tag, non-cloud tasks will be skipped.  On GCE this causes "Install tools" to be skipped, apparmor-utils is not installed, and then the "Enforcing ipsec with apparmor" step fails.

Also, fix the "reboot required by apt-get upgrade" check.

### DIFF
--- a/roles/common/tasks/main.yml
+++ b/roles/common/tasks/main.yml
@@ -2,9 +2,13 @@
 
 - name: Gather Facts
   setup:
+  tags:
+    - always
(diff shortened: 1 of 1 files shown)


dguido:
Can I get some review on this? @gunph1ld @defunctio

Issue #115, opened 2016-10-26
Display bug - "Congrats VPN setup" message appears in the middle of setup


ndtob:
The vpn : congrats message should appear at the very end of setup.

```

TASK [vpn : Iptables configured] ***********************************************
changed: [107.170.83.209] => (item={u'dest': u'/etc/iptables/rules.v4', u'src': u'rules.v4.j2'})
changed: [107.170.83.209] => (item={u'dest': u'/etc/iptables/rules.v6', u'src': u'rules.v6.j2'})

RUNNING HANDLER [vpn : restart strongswan] *************************************
changed: [107.170.83.209]

RUNNING HANDLER [vpn : restart iptables] ***************************************
changed: [107.170.83.209]

RUNNING HANDLER [vpn : congrats] ***********************************************
ok: [107.170.83.209] => {
    "msg": [
        "#----------------------------------------------------------------------#", 
        "#                          Congratulations!                            #", 
        "#                     Your Algo server is running.                     #", 
        "#    Config files and certificates are in the ./configs/ directory.    #", 
        "#          Go to https://www.dnsleaktest.com/ after connecting         #", 
        "#        and ensure that all your traffic passes through the VPN.      #", 
        "#          Local DNS resolver and Proxy IP address: 172.16.0.1", 
        "#----------------------------------------------------------------------#"
    ]
}

TASK [dns_adblocking : Gather Facts] *******************************************
ok: [107.170.83.209]

TASK [dns_adblocking : Dnsmasq installed] **************************************
changed: [107.170.83.209]

```

PR #130, opened 2016-11-26
Show congrats message at the end - #115

fkt:
Using a post_task seems to be a good idea here. The handler was called from a task in the vpn role and if other roles as dns_adblocking were added by the setup script the tasks of this role got executed after the handler.


dguido:
Looks good to me! Email dan@trailofbits.com to claim your bounty. Thanks!

fkt:
Please send it to a charity of your choice instead, if you like. :)

PR #156, opened 2016-12-12
Change default instance to free tier (t2.micro)

kennwhite:
I know this is a bit goofy, but the t2.nano is not in the free tier for AWS even though it is smaller than the t2.micro instance. See: https://aws.amazon.com/blogs/aws/ec2-update-t2-nano-instances-now-available/ (the "PS" at the bottom), confirmed on pricing page. The difference is $4.30 per mo vs. free/$8.76  per mo. Maybe add this to config questions, but at least one reviewer has noted this as an issue for his just-setup AWS free account.


defunctio:
fair enough

jpederzolli:
Does this make sense as a default? I see how it is better for brand new AWS users, but is seems better to default to the cheapest option and, as mentioned, add a new question about using the free-tier as part of deployment.

PR #172, opened 2016-12-19
Prep & additional requirements for RPM-based 6.x distros

kennwhite:
For non-systemd RPM-based distros (e.g., 6.8 Red Hat/CentOs, some Amazon Linux) the shipped python 2.6 libraries including pip and python-devel create a lot of issues for Ansible and Algo, particularly for critical libraries like pycrypto and Digital Ocean dopy. These pre-install steps allow end-to-end launch from RH/Cent 6.x VMs with no impact to core system functions like yum. 100% pass on shellcheck. Full (semi-redacted) build log launching against Digital Ocean is here: https://gist.github.com/kennwhite/20c520d251b87fb643beb38247033969


dguido:
Looks good! I'll clean this up a little bit and add a link from the readme later.

PR #171, opened 2016-12-19
Small UX hints

kennwhite:
Some small but hopefully useful UI improvements to help with new user setups.

PR #173, opened 2016-12-19
Revert "Small UX hints"

gunph1ld:
Reverts trailofbits/algo#171

- [ ] The key name should not be hardcoded.
- [ ] User is not root on Azure, GCE, EC2


kennwhite:
Should be addressed via PR: https://github.com/trailofbits/algo/pull/174

PR #174, opened 2016-12-20
Use Ansible variables for ssh key and user name in congrats text

kennwhite:
Addresses @gunph1ld PR: https://github.com/trailofbits/algo/pull/173. Patch verified via new deploys on both Digital Ocean and GCE instances. See: https://gist.github.com/kennwhite/c44f5144c6699e49f8ab3f3e568b0560


dguido:
Looks like it resolves the issues in #173

Issue #178, opened 2016-12-21
Deploys intermittently fail with timeout 


kennwhite:
### OS / Environment

Ubuntu 16.04.1 LTS on local VM

### Ansible version

ansible 2.2.0.0

### Version of components from `requirements.txt`

Name: ansible
Version: 2.2.0.0

Name: dopy
Version: 0.3.5

Name: boto
Version: 2.45.0

Name: azure
Version: 2.0.0rc5

Name: apache-libcloud
Version: 1.4.0

Name: six
Version: 1.10.0

Name: pyOpenSSL
Version: 16.2.0

### Summary of the problem

Roughly 1 out 3 deploys (with a 9 sample test) on GCE launches fails with a message: `"Timeout (12s) waiting for privilege escalation prompt: "` somewhere between `TASK [Install prerequisites]` and `TASK [common : Loopback for services configured]`. 

This has happened in the evenings, morning, and mid-afternoon with identical configs, and literally one launch would succeed, one would fail, two would succeed, one would fail, etc. 

### Steps to reproduce the behavior

1. All files in configs/* are removed except for GCE credential json file
2. (Using the same master trunk), attempt a launch on GCE
3. About every third launch times out somewhere between `TASK [Install prerequisites]` and `TASK [common : Loopback for services configured]`

### The way of deployment (cloud or local)

Ansible is on a local Ubuntu VM (Ubuntu 16.04.1 LTS), target is GCE

### Expected behavior

Existing time out values would suffice against nearly all GCE launches.

### Actual behavior

They do not.

### Full log

```
root@ubuntu:~/algo-2/algo-master# ./algo

  What provider would you like to use?
    1. DigitalOcean
    2. Amazon EC2
    3. Google Compute Engine
    4. Microsoft Azure
    5. Install to existing Ubuntu server

Enter the number of your desired provider
: 3

Enter the local path to your credentials JSON file (https://support.google.com/cloud/answer/6158849?hl=en&ref_topic=6262490#serviceaccounts):
: configs/gcp.json

Name the vpn server:
[algo]: algodns1221

  What zone should the server be located in?
    1. Central US       (Iowa A)
    2. Central US       (Iowa B)
    3. Central US       (Iowa C)
    4. Central US       (Iowa F)
    5. Eastern US       (South Carolina B)
    6. Eastern US       (South Carolina C)
    7. Eastern US       (South Carolina D)
    8. Western Europe   (Belgium B)
    9. Western Europe   (Belgium C)
[...]

Issue #146, opened 2016-12-11
Linux client support connectivity issues


defunctio:
### OS / Environment
Ubuntu 16.04 - strongswan 5.3.5

### Summary of the problem
Using the provided ipsec.conf configuration a connection can be established but no traffic traverses the VPN.

### Steps to reproduce the behavior
```
sudo apt-get install strongswan

root@ip-172-31-42-160:~/configs# cp 52.90.64.144_ca.crt /etc/ipsec.d/cacerts/
root@ip-172-31-42-160:~/configs# cp 52.90.64.144_dan.crt /etc/ipsec.d/certs/
root@ip-172-31-42-160:~/configs# cat 52.90.64.144_dan_ipsec.conf > /etc/ipsec.conf 
root@ip-172-31-42-160:~/configs# cat 52.90.64.144_dan_ipsec.secrets > /etc/ipsec.secrets 
root@ip-172-31-42-160:~/configs# cp 52.90.64.144_dan.key /etc/ipsec.d/private/
root@ip-172-31-42-160:~/configs# ipsec restart
root@ip-172-31-42-160:~/configs# ipsec up ikev2-52.90.64.144 
...
connection 'ikev2-52.90.64.144' established successfully

```

### Expected behavior
Traffic to be routed over VPN.

### Actual behavior
Not much...

I tested this both locally in a VM and on EC2 as a client, same results. Need to test to see if these results are reproducible for other people.

defunctio:
Compression should not be enabled by default. This is such a performance hit that many things just stop functioning properly. 

Remove the following line to resolve the issue.
https://github.com/trailofbits/algo/blob/master/roles/vpn/templates/client_ipsec.conf.j2#L6

douggard:
Same issue on 14.04 and the fix worked for me. A few sites still won't load (github, speedtest) but I can get some internet now.

defunctio:
@douggard try disabling fragmentation as well. If that doesn't help it  could also be related to MTU size. There needs to be further testing done on this subject though.

defunctio:
This does not resolve the underlying issue and compression should still not be enabled by default.

The default MTU size will likely be too large for some networks, especially on mobile networks.

For strongswan there are route based MTU/MSS settings `charon.plugins.kernel-netlink.mtu` & `charon.plugins.kernel-netlink.mtu` however it appears they may not work. I only did some initial testing and appears others were not able to get this feature to work properly as well.

[...]

PR #185, opened 2016-12-27
Drop the MSS for GCE instances

mutemule:
This forces the MSS for GCE instances only down to 1348, which seems to be the maximum value permitted for this tunnel. Without this, various parts of the internet appear to be offline through the VPN.

### Ansible version
```
ansible 2.2.0.0
  config file = /home/dwg/src/github.com/trailofbits/algo/ansible.cfg
  configured module search path = Default w/o overrides
```

```
Name: ansible
Version: 2.2.0.0
---
Name: dopy
Version: 0.3.5
---
Name: boto
Version: 2.45.0
---
Name: azure
Version: 2.0.0rc5
---
Name: apache-libcloud
Version: 1.4.0
---
Name: six
Version: 1.10.0
---
Name: pyOpenSSL
Version: 16.2.0
```

### Summary of the problem
Algo instances deployed in GCE don't have an appropriate MSS configured, so some sites -- notably `github.com` -- will seem to hang when attempting to load them, as packets marked with DF are being silently dropped.

1. Create a new `algo` instance in GCE.
2. Provision the VPN on an endpoint (any one; I've seen this on OS X, iOS, and Windows)
3. Attempt to load `https://github.com/`
4. Observe as it times out.

Bonus points: run a `tcpdump` locally and on the `algo` instance, and watch as packets arrive to the `algo` instance, but don't get forwarded across the tunnel.

### Expected behavior
The site loads.

### DIFF
--- a/roles/vpn/templates/rules.v4.j2
+++ b/roles/vpn/templates/rules.v4.j2
@@ -1,3 +1,13 @@
+*mangle
+:PREROUTING ACCEPT [0:0]
+:INPUT ACCEPT [0:0]
+:FORWARD ACCEPT [0:0]
+:OUTPUT ACCEPT [0:0]
+:POSTROUTING ACCEPT [0:0]
+{% if max_mss is defined %}
+-A FORWARD -s {{ vpn_network }} -p tcp -m tcp --tcp-flags SYN,RST SYN -j TCPMSS --set-mss {{ max_mss }}
+{% endif %}
+COMMIT
 *nat
 :PREROUTING ACCEPT [0:0]
 :POSTROUTING ACCEPT [0:0]
--- a/algo
+++ b/algo
@@ -286,7 +286,7 @@ Please choose the number of your zone. Press enter for default (#8) zone.
   esac
 
   ROLES="gce vpn cloud"
-  EXTRA_VARS="credentials_file=$credentials_file server_name=$server_name ssh_public_key=$ssh_public_key zone=$zone"
+  EXTRA_VARS="credentials_file=$credentials_file server_name=$server_name ssh_public_key=$ssh_public_key zone=$zone max_mss=1348"
 }
 
 non_cloud () {
--- a/roles/vpn/templates/rules.v6.j2
+++ b/roles/vpn/templates/rules.v6.j2
@@ -1,3 +1,13 @@
+*mangle
+:PREROUTING ACCEPT [0:0]
+:INPUT ACCEPT [0:0]
+:FORWARD ACCEPT [0:0]
+:OUTPUT ACCEPT [0:0]
+:POSTROUTING ACCEPT [0:0]
+{% if max_mss is defined %}
+-A FORWARD -s {{ vpn_network_ipv6 }} -p tcp -m tcp --tcp-flags SYN,RST SYN -j TCPMSS --set-mss {{ max_mss }}
+{% endif %}
+COMMIT
 *nat
 :PREROUTING ACCEPT [0:0]
 :POSTROUTING ACCEPT [0:0]


roles/vpn/templates/rules.v4.j2
```diff
@@ -1,3 +1,13 @@
+{% if ansible_product_name == 'Google Compute Engine' %}
+*mangle
+:PREROUTING ACCEPT [0:0]
+:INPUT ACCEPT [0:0]
+:FORWARD ACCEPT [0:0]
+:OUTPUT ACCEPT [0:0]
+:POSTROUTING ACCEPT [0:0]
+-A FORWARD -s {{ vpn_network }} -p tcp -m tcp --tcp-flags SYN,RST SYN -j TCPMSS --set-mss 1348
```
mutemule: As an alternative, an MSS value could be defined in the various `cloud-*` roles, then we could do away with the `ansible_product_name` conditional entirely.

gunph1ld:
This commit will break other cloud roles. `max_mss` should be defined for all the roles (as example you can use config.cfg)

defunctio:
See #146 also. This problem is likely not related to GCE and these options should really be set client side to prevent fragmentation in the first place when possible. Setting the MTU on my local interfaces / Android strongswan settings resolved issues with transmission failures and fragmentation for me.

mutemule:
> This commit will break other cloud roles. max_mss should be defined for all the roles (as example you can use config.cfg or default value for the variable)

It shouldn't. I tested this out in EC2, and it worked fine, and I also tested it by not defining `max_mss` for `cloud-gce`; the iptables rules files are deployed without issue, and the only difference is they don't have the MSS line in the `mangle` table.

>  This problem is likely not related to GCE

Correct. The algo instances I've created in EC2 work just fine, but I've not looked into it in great detail yet, so I wanted to keep this restricted to just GCE where I know the problem exists. My initial guess was that the MTU on EC2 instances is much, much higher than on GCE instances, but I didn't even look at client-side configuration.

>  these options should really be set client side to prevent fragmentation in the first place when possible

Agreed. I removed a paragraph from the initial PR pointing out that iptables is not a great place to do this, but it's the only place I can find that actually un-breaks things reliably (mostly because I can't adjust interface MTU in iOS). I'm well aware that this is really just a band-aid over the real problem, but it does give me a functional VPN.

defunctio:
Any chance someone can test this against other clients (OSX/Android/Linux/ChromeOS)? I'd like to merge this to resolve the immediate problem as long as it doesn't cause any conflict with other clients and open a new issue regarding the general issue. 

This is a problem across multiple providers and clients that should to be addressed promptly since it breaks functionality. We can move the changes to apply them globally or individually as we test different configurations to try to provide a balance between functionality for everyone and performance.

I can confirm that MTU settings that worked for me in the USA on T-mobile and ATT resulted in dropped packets on vodafone mobile while in Europe until the MTU was lowered further.

gunph1ld:
@mutemule Ah, I skipped the line `{% if max_mss is defined %}`. Sorry my blind spot :)

kennwhite:
This was a bit difficult to test, because I'm still hitting https://github.com/trailofbits/algo/issues/178 on GCE every other build or so (esp hanging on "Waiting for SSH..." and no instance seemed to be actually created). That said, I give this PR an enthusiastic thumbs up. Probably the fastest response I've seen (GCE or otherwise), and I threw every bloated thing I could think of at it: BusinessInsider, CNN, USAToday, LA Times, TMZ, even Tulsa World newspaper and the local Wichita Fox News. Ridiculously fast, no lags. Tested on OSX 10.11.6 (El Capitan) and iPhone 7 w/ latest iOS 10.2. I spent nearly 40 minutes of browsing, with no hesitation or timeouts.

Log attached in link below. Netflix fast.com clocked around 68 Mbps, but getting closer to ~200 Mbps on Ookla's beta.speedtest.net, which seems consistent with my web browsing performance.

[gce-mss-pr_log.txt](https://github.com/trailofbits/algo/files/678214/gce-mss-pr_log.txt)
screen shot 2016-12-29 at 10 04 40 pm

dguido:
Thanks for testing! This looks good to merge. Anyone else want to sign off on it?

dguido:
This looks fine. There may be other MTU problems with different cloud providers or intermediate networks (eg. AT&T) but at least this resolves one that we know about!

PR #209, opened 2017-01-05
EC2 encrypted volume support

defunctio:
This adds support for EC2 encrypted volumes.

Notes:
- algo.sh has **not** been updated to reflect the changes, I figure @gunph1ld can update the shell script to make these options visible easily for public usage.
- included `ec2_ami_copy` updated using boto3 for encrypted support
- ansible first checks the ./library folder before using system modules, so our modified module takes precedence
- tag inheritance is used to enable/disable encryption on EC2, to disable encryption `--skip-tags encrypted` must be applied
- `kms_key_id` extra variable has been added to add support for using the non default KMS
- If we have previously encrypted the Ubuntu AMI in this region, we will use the cached image first (this should have an age limit in the future + check if keys are different)

`ansible-playbook deploy.yml -t ec2,vpn,cloud -e "aws_server_name=algo ssh_public_key=~/.ssh/id_rsa.pub region=us-west-1 "`

`ansible-playbook deploy.yml -t ec2,vpn,cloud --skip-tags encrypted -e "aws_server_name=algo ssh_public_key=~/.ssh/id_rsa.pub region=us-west-1 "`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 ansible>=2.1
 dopy==0.3.5
-boto
-azure==2.0.0rc5
+boto>=2.5
+azure>=2.0.0rc5
 apache-libcloud
 six
 pyopenssl
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,8 @@
 ansible>=2.1
 dopy==0.3.5
 boto>=2.5
-azure>=2.0.0rc5
+boto3
+azure==2.0.0rc5
 apache-libcloud
 six
 pyopenssl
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -1,7 +1,7 @@
 ansible>=2.1
 dopy==0.3.5
-boto
-azure==2.0.0rc5
+boto>=2.5
```
gunph1ld: boto3 also should be here I think

requirements.txt
```diff
@@ -1,7 +1,7 @@
...
+azure>=2.0.0rc5
```
gunph1ld: should be [==](https://docs.ansible.com/ansible/guide_azure.html#requirements)

PR #211, opened 2017-01-10
Rename config.cfg to config.cfg.sample

sid77:
Hello, this PR renames `config.cfg` to `config.cfg.sample` and puts the original filename under `.gitignore`. The rationale behind this is to allow developers to hack away on algo and testing the changes without having to worry about the changes applied to the (now) checked in `config.cfg` file. I've also updated the documentation to reflect the extra `cp` needed to make everything work.


gunph1ld:
Makes life easier for developers, but not for users.

dguido:
Ah, I forgot to comment on this PR. Yes, as @gunph1ld said, we weighed the costs and benefits but we think that users deserve deference in this case. Developers should know enough to work around this issue. However, not all users will be ok with a series of new steps involving renaming or creating a new file.

We're open to other ideas if you have them!

sid77:
@dguido to be honest, I don't know! `¯\_(ツ)_/¯`

I've found the current setup quite puzzling: why do I need to add the `dan` and `jack` users to my VPN? Are they required? Well, _I_ know they're not required, but it's not easy to figure that out the first time if you aren't an experienced user.
From my perspective, the `.sample` extension clearly draw a line so the end user knows that the contents of the config file can be freely modified.

defunctio:
I agree that the default users should be removed and changed to just 'default'. Rather then providing an example cfg, comments should just serve as inline documentation.

PR #242, opened 2017-02-14
The azure tasks weren't pulling in credentials correctly from env

jacobwil:
If you weren't using the ~/.azure/credentials and the azure tasks ended up defaulting to environment variables the wrong values would be selected.


dguido:
Thanks!

Issue #246, opened 2017-02-17
Wrong filenames for files for some usernames with leading zeros in nicknames


lgg:
### OS / Environment
Ubuntu 16.04, Digital Ocean

### Ansible version
ansible 2.2.0.0

### Version of components from `requirements.txt`
dopy 0.3.5
boto  2.45.0
boto3 1.4.4
azure  2.0.0rc5
apache-libcloud 1.5.0
six 1.10.0
pyopenssl 16.1.0
jinja2 2.8

### Summary of the problem
See screenshots:
![image](https://cloud.githubusercontent.com/assets/12018713/23064305/64513080-f520-11e6-8e75-89ac06fa1e67.png)
![image](https://cloud.githubusercontent.com/assets/12018713/23064314/75e0ce28-f520-11e6-82e3-01cdd8d1ec7b.png)

For some reason some filenames are right: e.g. 009 and 029
and some are wrong: 001 became 1, 004 became 4 and etc

### Steps to reproduce the behavior
* edit config.cfg
* add users like in screenshot above
* create mobileconfigs

### The way of deployment (cloud or local)
* git clone this repo
* ./algo
* provide Digital Ocean api token

### Expected behavior
Mobileconfig files will have the same filenames as their usernames.

### Actual behavior
Some mobileconfigs lose leading zeros in their filenames

lgg:
The same thing is for all files: .mobileconfig, .p12 and others

gunph1ld:
Python interprets numbers that start with a zero as octal and converts it to decimal. Escape it

```
users:
  - '001'
  - '002'
  - '003'
```

PR #248, opened 2017-02-19
add info to README about escaping usernames

lgg:
It wasn't obvious for users how to escape nicknames and was it needed at all

see #246 for more

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,14 @@ The easiest way to get an Algo server running is to let it set up a _new_ virtua
 
  Linux (rpm-based): See the [Pre-Install Documentation for RedHat/CentOS 6.x](docs/pre-install_redhat_centos_6.x.md)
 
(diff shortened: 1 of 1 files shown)


dguido:
Hmm maybe we can explain this more concisely? It's very verbose considering the small number of users likely to make usernames with leading 0's.

dguido:
I think we can just say "(avoid usernames with leading 0's)" in parenthesis without explaining why when asking people to pick usernames.

lgg:
@dguido why users need to avoid them if they still can use them? 

In my case usernames with zeros exist because of SIP. (it's easier to tell digits instead of letters) And we have one system for all (in all services usernames and passwords are same).

jamesbroadhead:
Would this be better as a comment in the yaml file?

dguido:
If you want to make a discrete little comment on the yaml file, submit a PR please!

PR #251, opened 2017-02-20
Android client setup documentation

Lange:
This is a first draft of Android setup instructions. I expect that these will need some revising and that the formatting of this document will benefit from more tweaking.

We may also want to elaborate on the contents of the "advanced settings" menu ([screenshot](https://i.imgur.com/smsmdQF.png)).

The big chain of `&nbsp` characters in the table header is used to force the "Instructions" column to have a certain width, which then forces the "Screenshots" column to shrink down, which brings the embedded screenshots down to a more manageable size. Without this, the screenshots are huge and make the document almost unreadable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Find the corresponding mobileconfig (Apple Profile) for each user and send it to
 
 ### Android Devices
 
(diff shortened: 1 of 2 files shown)


Lange:
Ironically, this formatting hack looks terrible on phones. Need to continue playing with the formatting. Likely need to just render smaller thumbnails that link to full-size screenshots.

Lange:
This now looks much better on mobile devices ([screenshot](https://cloud.githubusercontent.com/assets/873012/23284740/bea040ae-f9f2-11e6-8778-656418294138.png))

PR #287, opened 2017-03-27
Add step in Windows instructions to install Six, clarify cloning repo

joshmeisels:
Based on my experience, if you go back to Step 4 in the general readme, it will fail due to Six not being installed. Added step to manually install Six.

Also added clarification on the command to clone the Algo repo.


dguido:
I think what you mentioned about Six is an issue that we can resolve. We'll take a closer look at it soon. I will add the `git clone` instruction immediately though!

dguido:
Can you file an issue regarding the Six issue if you haven't already? That will ensure we track and complete a fix for it.

Issue #299, opened 2017-03-29
Playbook fails with "NameError: name 'unicode' is not defined" on EC2 with Python 3.5/3.6 


williamsmj:
### OS / Environment

```
$ python --version
Python 3.5.2
```
or
```
$ python --version
Python 3.6.1
```

Tested on macOS

    Darwin 16.4.0 Darwin Kernel Version 16.4.0: Thu Dec 22 22:53:21 PST 2016; root:xnu-3789.41.3~3/RELEASE_X86_64 x86_64 i386 MacBookAir6,2 Darwin

### Ansible version

```
ansible 2.2.0.0
  config file = /Users/mike/tmp/algo/ansible.cfg
  configured module search path = Default w/o overrides
```

### Summary of the problem

Playbook exception with python 3.5.2 or 3.6.1. Goes away with python 2.7.13. Here's the exception.

```
 ________________________________________________________
/ TASK [cloud-ec2 : Locate official Ubuntu 16.04 AMI for \
\ region]                                                /
 --------------------------------------------------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

An exception occurred during task execution. To see the full traceback, use -vvv. The error was: NameError: name 'unicode' is not defined
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "module_stderr": "Traceback (most recent call last):\n  File \"/var/folders/cm/jf8_s09d53v3w4872py7jjlc0000gn/T/ansible_djclh5go/ansible_module_ec2_ami_find.py\", line 423, in <module>\n    main()\n  File \"/var/folders/cm/jf8_s09d53v3w4872py7jjlc0000gn/T/ansible_djclh5go/ansible_module_ec2_ami_find.py\", line 364, in main\n    ec2 = ec2_connect(module)\n  File \"/var/folders/cm/jf8_s09d53v3w4872py7jjlc0000gn/T/ansible_djclh5go/ansible_modlib.zip/ansible/module_utils/ec2.py\", line 272, in ec2_connect\n  File \"/var/folders/cm/jf8_s09d53v3w4872py7jjlc0000gn/T/ansible_djclh5go/ansible_modlib.zip/ansible/module_utils/ec2.py\", line 236, in get_aws_connection_info\nNameError: name 'unicode' is not defined\n", "module_stdout": "", "msg": "MODULE FAILURE"}
```

### Steps to reproduce the behavior

[...]

PR #302, opened 2017-03-29
Clarify python instructions

williamsmj:
- Be explicit about python2 requirement
- Remove `sudo` requirement on macOS
- Use `python -m pip` and `python -m virtualenv` to ensure expected python interpreter is running pip/virtualenv
- Tidy formatting

(See #299)


dguido:
This looks really good! Thanks @williamsmj.

PR #322, opened 2017-03-31
Check for creation of private key during its generation

caseylang:
This task was previously checking for the public key even though it is in place to generate the private key. A simple switch to the `creates` arg resolves the issue.


dguido:
Nice catch!

Issue #325, opened 2017-03-31
using public network for vpn network?


mankins:
### OS / Environment

OSX <-> Ubuntu

### Summary of the problem

A ``traceroute`` from the VPN client on the Mac shows:  

1  cpe-172-251-1-118.socal.res.rr.com (172.251.1.118)  168.280 ms  167.443 ms  167.360 ms

which doesn't look right?

``ifconfig`` on the client (Mac) shows that the local address:

```
	inet 172.20.0.178 netmask 0xfffffc00 broadcast 172.20.3.255
```

``ifconfig`` on the server (ubuntu, running on ec2):

```
eth0      Link encap:Ethernet  HWaddr 0e:07:c3:40:ad:52  
          inet addr:172.251.1.118  Bcast:172.251.1.255  Mask:255.255.255.0
```

From the ubuntu machine:

```
$ host 172.251.1.118
118.1.251.172.in-addr.arpa domain name pointer ip-172-251-1-118.ec2.internal.
```

From the mac:

```
$ host  172.251.1.118
118.1.251.172.in-addr.arpa domain name pointer cpe-172-251-1-118.socal.res.rr.com.
```

### Steps to reproduce the behavior

``traceroute mit.edu``
``ifconfig``
 
### The way of deployment (cloud or local)

cloud

### Expected behavior

IP should be actual cloud IP... or a [private network IP](https://en.wikipedia.org/wiki/Private_network).

I think that the network size was imagined to be larger, but is actually just 172.16.0.0 – 172.31.255.255

### Actual behavior

Uses someone else's public IP. In this case ``172.251.1.118``

PR #330, opened 2017-03-31
Change EC2 VPC CIDR blocks to non-routable addresses.

joshwatson:
The previous address ranges were actually routable addresses, which caused some concern for some people because it looked suspicious in tracert. The new CIDR blocks are non-routable addresses, which resolves this concern.

(See #325)

### DIFF
--- a/roles/cloud-ec2/defaults/main.yml
+++ b/roles/cloud-ec2/defaults/main.yml
@@ -1,5 +1,5 @@
 ---
 
 ec2_vpc_nets:
-  cidr_block: 172.251.0.0/23
-  subnet_cidr: 172.251.1.0/24
+  cidr_block: 192.168.0.0/23
+  subnet_cidr: 192.168.1.0/24


mankins:
I wonder if 192.168.0.0/23 etc are too popular as client endpoints to make them useful in this capacity. I'd think just fixing the range to be private instead of public would be better? I mean, how many people have 192.168.0.1 as their router? I'd think many?

172.16.0.0/12

dguido:
Ah yes! Can you pick a good range that is unlikely to collide with people's personal routers and submit a PR?

defunctio:
This is actually auto assigned for EC2 VPCs when they are not declared. I'm not sure why this was declared, it shouldn't be required but it may have been added for some edge case.

dguido:
That's odd, can some try it without the declaration and tell me if it works?

mankins:
I just submitted https://github.com/trailofbits/algo/pull/335 ... not sure about the declaration. I did a little searching and didn't totally grok what was going on, but figured the cidr shift was progress enough.

Issue #291, opened 2017-03-28
Autopopulate list of zones/regions using a more authoritative list.


tariqajyusuf:
Some of the cloud providers list the available zones via an API. Using these can save the trouble of updating whenever new zones open up (See #286). We already get the API keys required to make this happen, we just need to change the order and grep the responses accordingly:

* **Google Cloud**: https://cloud.google.com/compute/docs/reference/latest/zones/list
* **DigitalOcean**: https://developers.digitalocean.com/documentation/v2/#list-all-regions
* **AWS EC2**: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRegions.html
* **Azure**: https://docs.microsoft.com/en-us/rest/api/apimanagement/regions

jackivanov:
https://github.com/trailofbits/algo/issues/941

PR #332, opened 2017-04-01
Add new Azure regions and allow user to select VM size

joshmeisels:
- Added a bunch of additional Azure regions
- Added a prompt to let the user choose VM size between 3 options, including 2 super low cost ones.
- Removed the default VM size of D1 which is being phased out, replaced with D1v2.

Future enhancements could include:
- Asking for a continent before the Azure region - this way it isn't such a long list to choose from.
- Auto-pulling of Azure regions as described in https://github.com/trailofbits/algo/issues/291

### DIFF
--- a/algo
+++ b/algo
@@ -157,19 +157,17 @@ Enter the number of your desired region:
 
   read -p "
   What size do you want your VM?
-  1) Standard D1 (not available in all regions)
-  2) Standard D1 v2
-  3) Standard A0 (cheap)
-  4) Basic A0 (cheapest)
+  1) Standard D1 v2
+  2) Standard A0 (cheap)
+  3) Basic A0 (cheapest)
 Enter the number of your desired size
 [1]: " -r azure_vm_size
   azure_vm_size=${azure_vm_size:-1}
   
   case "$azure_vm_size" in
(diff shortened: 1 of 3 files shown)


dguido:
Thanks for this PR! Can you remove the bits about selecting a VM size? I'd like to move any configuration of VM size into the config.cfg file to avoid prompting users about it. Then, advanced users can dig into the config file to edit it if necessary. The default size should work for 99% of all use cases. Each new question we add to the initial script adds to the complexity for regular users.

joshmeisels:
@dguido sure, I'll remove that part. Should the default size be `Standard_D1v2` or `Basic_A0`?

Per my post on Slack, D1v2 is way more expensive, but the performance suffers on Basic_A0. I assume it should default to the more expensive size with better perf.

dguido:
Ah, you assume wrong! A0 is more in line with our philosophy of using the least resources that will suffice. Our user base is also cost sensitive.

joshmeisels:
Thanks @dguido. I've undone the change to let the user select a VM size and changed the hard coded value to the cheapest one.

dguido:
Thanks!

PR #335, opened 2017-04-02
Change EC2 VPC CIDR blocks to uncommon non-routable addresses

mankins:
While the last PR keeps the ips in the private range, it will look show 192.168.x.x in traceroutes as well, which will confuse some people, I suspect.

This fixes https://github.com/trailofbits/algo/issues/325

### DIFF
--- a/roles/cloud-ec2/defaults/main.yml
+++ b/roles/cloud-ec2/defaults/main.yml
@@ -1,5 +1,5 @@
 ---
 
 ec2_vpc_nets:
-  cidr_block: 192.168.0.0/23
-  subnet_cidr: 192.168.1.0/24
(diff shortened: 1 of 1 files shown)


dguido:
What happens if we omit listing them at all?

defunctio:
I'm fairly certain the default VPC will be used, it may have been added to cover the edge cause of no default VPC existing in a region that has never been used before. Looks like @gunph1ld added it [awhile back](https://github.com/trailofbits/algo/blame/master/roles/cloud-ec2/tasks/main.yml#L66 ) so I can't confirm that was why. There is no harm but it probably should be conditional.

dguido:
ok an uncommon VPC address will do then. Pick one

Issue #288, opened 2017-03-27
Add URLProbe for iOS configuration


tariqajyusuf:
I came across an issue with the iOS configuration where I was unable to connect through to most captive portals due to failing VPN connections. Whenever shutting the VPN off, it immediately turns back on.

It looks like this can be [solved](https://www.derman.com/blogs/Example-iOS-VPN-OnDemand-Rules) with adding a `URLStringProbe` as one of the `OnDemandRuless`. Happy to take this and see if I can get it working.

An idea for rules would be:

* Check if Proxy IP is accessible
* Check if google.com is accessible
* If not, disable proxy.

dguido:
Neat! Yes, it would be great if you could try setting up a mobileconfig profile with these options and let us know the result. Our lead developer is on vacation for a little bit :-P.

tariqajyusuf:
Will do, I'm currently away from my main dev box but I'll try to get a proof of concept next week. If all goes well I'll let you know and try to get it together in a pull request.

tariqajyusuf:
I've created a pull request for this issue. It looks like I'm only able to probe an HTTP URL so I was only able to do the `google.com` check mentioned here.

PR #336, opened 2017-04-02
Add `URLStringProbe` to iOS config to all networks to allow access to captive portals

tariqajyusuf:
See Feature #288 

Currently, the behavior on iOS is to try to connect to VPN regardless of if the internet is accessible or not. When disabling a VPN for a captive portal scenario, it disables the VPN and disconnects the current network. This causes it to switch to another network (either a wifi network or mobile network) triggering an OnDemand reconnect. This rule disables VPN until the internet is available (using `google.com`) as the endpoint.

I need to validate that the VPN will autoreconnect as soon as google.com is available, but this is currently working better than the current solution. I've tested that the reconnection attempt doesn't happen on a captive network that hasn't been connected yet. `xfinitywifi` has a captive portal that you need to log in to.

**Current configuration generated by HEAD**
img_1503

**Configuration generated by this version**
img_1504

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2016 Trail of Bits
+Copyright (c) 2016 Trail of Bits, Google Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/roles/vpn/templates/mobileconfig.j2
+++ b/roles/vpn/templates/mobileconfig.j2
@@ -37,6 +37,8 @@
 {% endif %}
                     <key>InterfaceTypeMatch</key>
                       <string>WiFi</string>
+                    <key>URLStringProbe</key>
+                      <string>http://www.gstatic.com/generate_204</string>
                   </dict>
                   <dict>
                     <key>Action</key>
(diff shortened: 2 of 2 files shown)


tariqajyusuf:
I need to check with my employers open source policy to make sure that I'm allowed to contribute but otherwise this commit is solid.

gunph1ld:
According to the Apple docs:
> A URL to probe. If this URL is successfully fetched (returning a 200 HTTP status code) without redirection, this rule matches.

```
$ curl -I http://www.google.com
HTTP/1.1 302 Found
```

We need to chose the url more properly I guess

tariqajyusuf:
I just remembered we can use gstatic.com/generate_204. That's used by
Android to probe for network connections.

dguido:
Can you update this rule with the gststatic address?

tariqajyusuf:
Updated to use gstatic URL

dguido:
This looks good but you shouldn't modify the license.

tariqajyusuf:
My employer requested it since I'm committing as an employee.

dguido:
Can you explain more? Who at Google told you to modify our license file? I've never seen this before.

dguido:
Welp, we can't accept this pull request. Sorry! Let me know if you find out more info about open source licensing that would help us understand the issue.

PR #343, opened 2017-04-03
Minor AWS improvements

pc-0:
This pull request has two minor changes for AWS:

1. Check for standard AWS environment variables before asking the user. The variables `$AWS_ACCESS_KEY_ID`, `$AWS_SECRET_ACCESS_KEY`, and `$AWS_DEFAULT_REGION` are all standards used by tools like the [aws-cli](https://github.com/aws/aws-cli).
1. Use an [Elastic IP Address](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/elastic-ip-addresses-eip.html) on the VPN server instead of a regular public IP. Elastic IPs are free to use, persist beyond an instance stop, and can be attached/detached to a different instance.

### DIFF
--- a/algo
+++ b/algo
@@ -208,59 +208,70 @@ EXTRA_VARS="do_access_token=$do_access_token do_server_name=$do_server_name do_r
 }
 
 ec2 () {
-  read -p "
-Enter your aws_access_key (http://docs.aws.amazon.com/general/latest/gr/managing-aws-access-keys.html)
-Note: Make sure to use either your root key (recommended) or an IAM user with an acceptable policy attached
-[pasted values will not be displayed]
-[AKIA...]: " -rs aws_access_key
+  if [[ -z $AWS_ACCESS_KEY_ID ]]; then
+    read -p "
+  Enter your aws_access_key (http://docs.aws.amazon.com/general/latest/gr/managing-aws-access-keys.html)
+  Note: Make sure to use either your root key (recommended) or an IAM user with an acceptable policy attached
+  [pasted values will not be displayed]
+  [AKIA...]: " -rs aws_access_key
+  else
+    aws_access_key=$AWS_ACCESS_KEY_ID
+  fi
+
+  if [[ -z $AWS_SECRET_ACCESS_KEY ]]; then
+    read -p "
+  Enter your aws_secret_key (http://docs.aws.amazon.com/general/latest/gr/managing-aws-access-keys.html)
+  Note: Make sure to use either your root key (recommended) or an IAM user with an acceptable policy attached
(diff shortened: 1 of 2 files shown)


algo
```diff
@@ -208,59 +208,70 @@ EXTRA_VARS="do_access_token=$do_access_token do_server_name=$do_server_name do_r
 }
 
 ec2 () {
-  read -p "
-Enter your aws_access_key (http://docs.aws.amazon.com/general/latest/gr/managing-aws-access-keys.html)
-Note: Make sure to use either your root key (recommended) or an IAM user with an acceptable policy attached
-[pasted values will not be displayed]
-[AKIA...]: " -rs aws_access_key
+  if [[ -z $AWS_ACCESS_KEY_ID ]]; then
```
defunctio: These are already implied here.
https://github.com/trailofbits/algo/blob/master/roles/cloud-ec2/tasks/main.yml#L3

algo
```diff
@@ -208,59 +208,70 @@ EXTRA_VARS="do_access_token=$do_access_token do_server_name=$do_server_name do_r
...
+    read -p "
+  Enter your aws_access_key (http://docs.aws.amazon.com/general/latest/gr/managing-aws-access-keys.html)
+  Note: Make sure to use either your root key (recommended) or an IAM user with an acceptable policy attached
+  [pasted values will not be displayed]
+  [AKIA...]: " -rs aws_access_key
+  else
+    aws_access_key=$AWS_ACCESS_KEY_ID
+  fi
+
+  if [[ -z $AWS_SECRET_ACCESS_KEY ]]; then
```
defunctio: Same as above.

algo
```diff
@@ -208,59 +208,70 @@ EXTRA_VARS="do_access_token=$do_access_token do_server_name=$do_server_name do_r
...
-    11) region="eu-west-1" ;;
-	12) region="eu-west-2";;
-    13) region="sa-east-1" ;;
-    14) region="ca-central-1" ;;
-  esac
+  Name the vpn server:
+  [algo]: " -r aws_server_name
+    aws_server_name=${aws_server_name:-algo}
+
+  if [[ -z $AWS_DEFAULT_REGION ]]; then
```
defunctio: Let's move this into roles/cloud-ec2/tasks/main.yml in the same format as the keys. I would like to keep the env lookups in the roles themselves rather than the bash script so that there is no dependency on the script for the roles to function.

defunctio:
@pc-0 please note the requested changes on your PR.

defunctio:
Closing due to lack of response from PR owner.

PR #350, opened 2017-04-04
Update Powershell Instructions for VPN connection

### LINKED ISSUE #329: Windows Powershell script fails to install cert due to permissions
### OS / Environment
Windows 10  - CLIENT setup

### Summary of the problem
Running the user powershell script created for Windows users does not install the user cert because this action requires elevation. The script should be updated to elevate before attempting to install the cert.

Set up a VPN server using Algo. Copy the Powershell script and cert files to the windows machine you want to join the VPN. Run the Powershell script.

### Expected behavior
Script should install cert and create the VPN connection in Windows.

### Actual behavior
Script shows an error on installing the cert due to lack of permissions.

xtrapwr:
Update the Powershell instructions for creating the Windows VPN
connection.  Clarified requirement that Powershell run as Administrator,
and additional steps required to update the Execution Policies for
PowerShell. Resolve #329

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ You need to install the [strongSwan VPN Client for Android 4 and newer](https://
 ### Windows
 
 Copy the CA certificate, user certificate, and the user PowerShell script to the client computer. Import the CA certificate to the local machine Trusted Root certificate store. Then, run the included PowerShell script to import the user certificate, set up a VPN connection, and activate stronger ciphers on it. 
-The PowerShell script has to be run as Administrator, so first open Powershell as Administrator, then navigate to your copied files. If you have never used PowerShell before, you will need to change the Execution Policy to allow unsigned scripts to run. Execute the following command in PowerShell to do so.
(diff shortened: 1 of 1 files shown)


dguido:
Windows is so ugly :-x.

Can you put the content you added in a new paragraph (add a newline)? Also, standardize on a single capitalization of PowerShell please. We only use one space after a period.

Also, clarification for me... do some people NOT need these commands? What is the default behavior of Windows? Is this only an issue when you're using a corporate managed computer or running as a lower-privileged user? I don't use Windows so I'm just trying to get some context here.

xtrapwr:
Windows is indeed very ugly. I'll do some formatting niceness and resubmit the Pull.

To clarify, on a default installation of Windows 10 (all versions) the PowerShell Execution Policy is Restricted. Most developers would have changed that on their own as part of setting up PowerShell, but I expect the layperson would not. I tested on a clean build of Windows 10 Home, but I suspect it is the same for Pro and Enterprise, as part of their goal to be more secure by default.

dguido:
Got it. Thanks for the extra info!

xtrapwr:
Pull should now incorporate all the formatting fixes requested.

climatefixer:
I am new to this site and want to learn how to use the VPN when sending emails from my tutanota.com account.  climatefixer@tutanota.com

PR #363, opened 2017-04-06
Modify creation of GCE Instance

caseylang:
This commit adjusts the creation of the GCE instance, specifically related to the addition of the `sshKeys` metadata. That key has been deprecated in favor of `ssh-keys` ([reference](https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys)).

Also changed is the association of the ssh key with the root user, which is by default disabled on GCP public images ([reference](https://cloud.google.com/compute/docs/instances/ssh-keys#root)).

### DIFF
--- a/roles/cloud-gce/tasks/main.yml
+++ b/roles/cloud-gce/tasks/main.yml
@@ -15,10 +15,10 @@
     zone: "{{ zone }}"
     machine_type: f1-micro
     image: ubuntu-1604
-    service_account_email: "{{ service_account_email  }}"
-    credentials_file: "{{ credentials_file_path  }}"
-    project_id: "{{ project_id  }}"
-    metadata: '{"sshKeys":"root:{{ ssh_public_key_lookup }}"}'
+    service_account_email: "{{ service_account_email }}"
+    credentials_file: "{{ credentials_file_path }}"
+    project_id: "{{ project_id }}"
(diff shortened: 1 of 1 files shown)


roles/cloud-gce/tasks/main.yml
```diff
@@ -15,10 +15,10 @@
...
     machine_type: f1-micro
     image: ubuntu-1604
-    service_account_email: "{{ service_account_email  }}"
-    credentials_file: "{{ credentials_file_path  }}"
-    project_id: "{{ project_id  }}"
-    metadata: '{"sshKeys":"root:{{ ssh_public_key_lookup }}"}'
+    service_account_email: "{{ service_account_email }}"
+    credentials_file: "{{ credentials_file_path }}"
+    project_id: "{{ project_id }}"
+    metadata: '{"ssh-keys":"algo:{{ ssh_public_key_lookup }}"}'
```
defunctio: For the sake of uniformity across all Ubuntu deployments let's use the default `ubuntu` user.

roles/cloud-gce/tasks/main.yml
```diff
@@ -27,7 +27,7 @@
   add_host:
     name: "{{ google_vm.instance_data[0].public_ip }}"
     groups: vpn-host
-    ansible_ssh_user: ubuntu
+    ansible_ssh_user: algo
```
defunctio: as noted above

caseylang:
Requested changes complete

Issue #123, opened 2016-10-30
EC2: IAM inline policy


defunctio:
Provide an IAM inline policy for users to limit only the permissions Algo will need on AWS for deployment rather than granting `AmazonEC2FullAccess`

dguido:
$100 bounty! Submit a pull request and email dan@trailofbits.com to claim it. Partial solutions may be rewarded.

defunctio:
This is a start to an inline IAM policy that 'should' include all the necessary actions for Algo to perform the tasks it needs for deployment. However; additional resource restrictions should be implemented on actions that allow for it to limit the policy as much as possible. 

Note: Certain actions such as `CreateKeyPair`, `DeleteKeyPair`, `ImportKeyPair` and many others have no resource limitation filters.

I will not be claiming any bounties so feel free to use this partial solution to create a more complete solution and submit your PR to claim your bounty!

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "ec2:DescribeInstances",
                "ec2:DescribeImages",
                "ec2:DescribeKeyPairs",
                "ec2:DescribeVpcs",
                "ec2:DescribeSubnets",
                "ec2:DescribeSecurityGroups",
                "ec2:CreateSecurityGroup",
                "ec2:AuthorizeSecurityGroupEgress",
                "ec2:AuthorizeSecurityGroupIngress",
                "ec2:CreateTags",
                "ec2:CreateKeyPair",
                "ec2:DeleteKeyPair",
                "ec2:ImportKeyPair"
            ],
            "Resource": "*"
        },
        {
            "Effect": "Allow",
            "Action": "ec2:RunInstances",
            "Resource": "*"
        }
    ]
}
```

PR #364, opened 2017-04-07
Created inline policy that limits actions, where supported, to resour…

spankminister:
…ces tagged with Environment:Algo

### DIFF
--- a/library/inlinepolicy.json
+++ b/library/inlinepolicy.json
@@ -1,0 +1,54 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
+                "ec2:AssociateRouteTable",
+                "ec2:AttachInternetGateway",
+                "ec2:AuthorizeSecurityGroupEgress",
+                "ec2:AuthorizeSecurityGroupIngress",
+                "ec2:CreateSubnet",
+                "ec2:CreateTags",
+                "ec2:CreateVpc",
+                "ec2:CreateSubnet",
+                "ec2:CreateRoute",
+                "ec2:CreateRouteTable",
+                "ec2:CreateInternetGateway",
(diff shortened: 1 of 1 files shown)


defunctio:
Referencing issue #123 as there was a bounty on this issue.

dguido:
How do we "hook this up" to Algo? Is this something we put in the docs, something we use on their API? I'm not very familiar with IAM.

defunctio:
The ACL needs to be tested first but it's usage will just need to be added to advanced usage documentation. It's something the user will need to configure manually on AWS to use.

spankminister:
I noticed the PR has been marked "Needs Testing," should I write the test also? I would just write up a test where I create an instance with a user that has all permissions, then ensure that the Algo user with the inline policy can't modify the untagged VM in any of the ways specified.

I made the policy by running the ansible-playbook command and adding a permission based on the error for which EC2 action failed. When I got the minimum number of permissions for Algo to succeed, I then moved some of those actions and resource constrained them based on the [list of supported APIs on this page](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-supported-iam-actions-resources.html).

defunctio:
It just means that we need to manually test it on our end before we merge it, I just haven't tested it yet.

If you could add documentation and add AWS reference material in the advanced doc section that detail how users can setup IAM users, assign ACL permissions and generate an API key to use with Algo that would also be helpful.

aboutte:
It would also be nice if these two lines of code were updated to reference the new documentation:

https://github.com/trailofbits/algo/blob/master/algo#L222
https://github.com/trailofbits/algo/blob/master/algo#L229

Not only is it not recommended to use the root key and secret it is recommended to NOT generate a key and secret for the root user.  :)

defunctio:
Closing this pull request in favor of #447 for CF support that will be merged soon.

Issue #263, opened 2017-03-03
Strongswan networkmanager plugin - Ubuntu 16.04


defunctio:
Networkmanager has a UI plugin for strongswan, however the version of the package that ships with Ubuntu 16.04 does not function properly (does not appear in the menu). It has since been resolved in the source tree but the package for Ubuntu has not been updated. 

https://wiki.strongswan.org/issues/1429
`Fixed with version 1.4.0 of the plugin.`
```
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Harald Dunkel <harri@afaics.de>
Architecture: amd64
Version: 1.3.1-1ubuntu1
```

Another note is that nmcli does have the ability to import configs/profiles on some NM plugins (openvpn, etc); however this feature is not implemented in the strongswan module and last I checked there was no issue on the matter.

conorsch:
> Another note is that nmcli does have the ability to import configs/profiles on some NM plugins (openvpn, etc)

I'm able to run `nmcli con import type openvpn file <username>.ovpn` just fine here, on nmcli v1.6.2 under Debian Stretch. The Ubuntu package may not provide this functionality, of course; haven't checked there.

rufoa:
Neither the n-m applet nor `nmcli` supports importing strongswan configs fwiw (unlike the ovpn plugin):
```
$ nmcli con import type strongswan file rufo.sswan 
Error: failed to import 'rufo.sswan': the plugin does not support import capability.
```
There is no way to provide a custom cipher suite etc using the n-m plugin, which means negotiation fails with the default Algo server config. If you enable weak crypto ("Windows 10 support") in Algo it succeeds, but this isn't ideal.

adworacz:
I just ran into this yesterday myself, and spent 20 minutes searching the web trying to figure out what was going on.

As it stands, we can setup an Ubuntu 16.04 server with Algo, but we can't use Ubuntu 16.04 as a client, which is an unfortunately poor user experience.

It is likely worth adding a line to the README/documentation somewhere noting that this is a known issue, and it should be fixed with the release of 17.04 (I believe).

roycewilliams:
The README has been updated to note that Ubuntu 17.04 is the minimum version supported, but a little more info about why could be useful.

defunctio:
[...]

PR #372, opened 2017-04-09
Support for Ubuntu network-manager VPN connections

defunctio:
Do we want to make yet another condition for cipher selection for Ubuntu or just support this unconditionally? I've just added it to the cipher list for now. 

Looks like libcharon doesn't enable any AES GCM ciphers by default and opts to enable Camellia CCM ciphers instead (export restrictions? idk). 

See #263


dguido:
Let's add this to the new ciphers branch. Instead of "windows" we can call it "compatible". Then "compatible" can support both Windows and Ubuntu w/ Network Manager. IMHO go for aes128, the extra headroom up to 256 is worthless.

defunctio:
Sounds good to me.

Issue #132, opened 2016-11-26
EC2 AWS: Cloudformation


defunctio:
In a project to implement Algo as a native mobile application to both iOS and Android it was necessary to migrate much of the AWS infrastructure deployment for Algo into a [Cloudformation](https://aws.amazon.com/cloudformation/) template to eliminate dependencies, simplify development and reduce the required infrastructure to deploy Algo as a mobile application.

There are a few benefits over using Cloudformation on AWS rather than programmatically managing the resources necessary for Algo as we are currently doing.

- All resources are created, owned and managed by the stack owner.
- All resources are unique named to prevent naming collisions and promote a multi-user environment.
- Better resource limit management and cost estimation.
- Destroying a stack removes ALL resources associated with that instance of Algo.

In short, it simplifies managing all the necessary resources for the end user and prevents mistakes. It may be worth looking into the plausibility of reusing the template for Algo itself in the future.

PR #385, opened 2017-04-10
Allocate an EIP in AWS

anderm3:
based on pc-0's PR https://github.com/trailofbits/algo/pull/343

### DIFF
--- a/roles/cloud-ec2/tasks/main.yml
+++ b/roles/cloud-ec2/tasks/main.yml
@@ -104,9 +104,20 @@
     instance_initiated_shutdown_behavior: terminate
   register: ec2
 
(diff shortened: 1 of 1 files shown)


gunph1ld:
From #394
> You only pay for the Elastic IP when it's not attached to an instance.

I'm pretty sure, users will forgot to delete it after terminating an instance. 
And my opinion is we need to totally move to Cloudformation #132

defunctio:
We do want to move to Cloudformation for better resource management as noted above by @gunph1ld . Also I think using an EIP should be an advanced optional setting IMO, not a default. I'm going to reject the PR and let's use #394 to track the issue.

PR #412, opened 2017-04-15
Only enable ChaCha cipher

jauderho:
See https://stribika.github.io/2015/01/04/secure-secure-shell.html for some good ideas.

I would not enable ECDSA. Most up to date clients should now support chacha/ed25519. If need be, we can add back some RSA as fallback.

### DIFF
--- a/roles/security/templates/sshd_config.j2
+++ b/roles/security/templates/sshd_config.j2
@@ -41,9 +41,12 @@ PrintMotd no
 PrintLastLog yes
 
 # Use only modern host keys
+HostKey /etc/ssh/ssh_host_ecdsa_key
 HostKey /etc/ssh/ssh_host_ed25519_key
 
 # Use only modern ciphers
-KexAlgorithms curve25519-sha256@libssh.org
-Ciphers chacha20-poly1305@openssh.com
-MACs hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com
+KexAlgorithms curve25519-sha256@libssh.org,ecdh-sha2-nistp256
+Ciphers chacha20-poly1305@openssh.com,aes128-gcm@openssh.com
+MACs hmac-sha2-256-etm@openssh.com
+HostKeyAlgorithms ssh-ed25519,ecdsa-sha2-nistp256
+# PubkeyAcceptedKeyTypes accept anything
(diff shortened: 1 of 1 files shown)


dguido:
This seems reasonable. We could even drop the last `hmac-sha2-256-etm@openssh.com`.

Is there any reason this won't work in DigitalOcean, AWS, or Azure? We're stuck with RSA private keys due to AWS but I don't think these settings will interfere with that.

jauderho:
Someone will have to test on those providers, I have not gotten around to getting access/accounts yet. 

I noticed the reversion to RSA keys after I submitted this request and wasn't sure what the reason was. 

As mentioned, it's entirely reasonable to add RSA as fallback. Based on my reading of the config files, I would think there is an option to create ed25519 and RSA keys giving preference to the former.

One could then end up with a number of identity files to offer up. For example, I have:

> Host *
> IdentityFile ~/.ssh/id_ed25519
> IdentityFile ~/.ssh/id_ra

Which allows for graceful fallback.

pguizeline:
Using only ed25519 would break iPhone ssh/sftp clients, not that it really matters, but having ECDSA or RSA as a fallback would be really nice. I really don't know if RSA 4096 is better or worse than ECDSA 521.

As for AWS it would be necessary to provision a temporary RSA key to deploy the instance, and then change the public key to ed25519 afterwards.

dguido:
I think you should replace it with this:

```
# Use only modern host keys
HostKey /etc/ssh/ssh_host_ecdsa_key
HostKey /etc/ssh/ssh_host_ed25519_key

# Use only modern ciphers
KexAlgorithms curve25519-sha256@libssh.org,ecdh-sha2-nistp256
Ciphers chacha20-poly1305@openssh.com,aes128-gcm@openssh.com
MACs hmac-sha2-256-etm@openssh.com
HostKeyAlgorithms ssh-ed25519-cert-v01@openssh.com,ecdsa-sha2-nistp256-cert-v01@openssh.com
# PubkeyAcceptedKeyTypes accept anything
```

I just tested and this works on AWS. It is less than before. We need PubkeyAcceptedKeyTypes to remain open since AWS requires RSA pubkeys. The others we are free to play with so I simply stuck with the best Bernstein curve and a solid NIST curve as a second option.

PR #416, opened 2017-04-15
check do_access_token before adding it to EXTRA_VARS

aboutte:


gunph1ld:
Could you provide more explanation? Why do we need this?

gunph1ld:
There is [another way ](https://github.com/trailofbits/algo/pull/426)

PR #419, opened 2017-04-16
Improve Ubuntu Instructions

logannc:
Added left out steps and a common pitfall (lack of openssl plugin), removing the initial Linux instructions as they no longer add much value.


dguido:
Thanks! I'm going to accept this but add a little bit of context back.

PR #423, opened 2017-04-16
Ensure root .ssh directory exists

rossey:
The .ssh directory doesn't exist with a fresh install of Ubuntu 16.04.


gunph1ld:
I'm not sure why we use such ways, but we need to replace some tasks (lines 51-64 and 71-72) to local_action.

rossey:
Thanks for taking a look @gunph1ld. Does bf75a1bb035d20cf478e8df1d9b6fccbfaef8411 fix this?

gunph1ld:
@rossey Yes, it does

PR #427, opened 2017-04-16
Update troubleshooting with note about ip frag

MiWCryptAnalytics:
note about ip fragmentation on consumer routers


dguido:
Can you add a little more detail for users who probably will not use Wireshark to debug their own connections? And what does "Try disabling any firewall settings on the router related to blocking or dropping IP fragments" typically look like? What did it look like on your router?

dguido:
I got it :-P. Thanks!

PR #428, opened 2017-04-16
Add URLStringProbe

### LINKED ISSUE #288: Add URLProbe for iOS configuration

gunph1ld:
Fixes #288

I'm not sure about the url we use. According to the Apple docs, we have to get a 200 HTTP.

### DIFF
--- a/roles/vpn/templates/mobileconfig.j2
+++ b/roles/vpn/templates/mobileconfig.j2
@@ -37,6 +37,8 @@
 {% endif %}
                     <key>InterfaceTypeMatch</key>
                       <string>WiFi</string>
+                    <key>URLStringProbe</key>
+                      <string>http://www.gstatic.com/generate_204</string>
                   </dict>
                   <dict>
                     <key>Action</key>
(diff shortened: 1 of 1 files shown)


gunph1ld:
Yes, working

pguizeline:
I've just did my second test using DO deployment. I've manually checked the .mobileconfig to see if the the URLStringProbe changes were present, and they were. First test was on EC2. Both deployments got the same results:

Algo is running with no problems on the latest iOS 10.3.1, both on 4G and Wi-Fi

dguido:
Let's use http://captive.apple.com/hotspot-detect.html instead since it gives a 200 response, not a 204, which is what the docs recommend.

Issue #60, opened 2016-08-18
IPv6 for EC2


dguido:
We may want to remove the playbook steps that refer to IPv6 on EC2 and GCE, unless we can get IPv6 working on those clouds.

defunctio:
I can confirm that AWS does not support IPv6 on EC2.

gunph1ld:
> Google Compute Engine networks do not support IPv6 at all.

gunph1ld:
IPv6 enabled only for Digital Ocean, yes, but we need to configure IPv6 local network and a default gateway for a VPN connection, to avoid passing traffic to an ISP network (if an ISP supports IPv6)

defunctio:
https://aws.amazon.com/blogs/aws/new-ipv6-support-for-ec2-instances-in-virtual-private-clouds/

dguido:
https://wiki.strongswan.org/issues/817#note-4

samkelleher:
Just to loop in, AWS EC2 now has IPv6 in [most of its regions](https://aws.amazon.com/blogs/aws/aws-ipv6-update-global-support-spanning-15-regions-multiple-aws-services/).

aboutte:
When #431 is merged this can be closed.

PR #431, opened 2017-04-17
CF tested and working for EC2 deployment

### LINKED ISSUE #132: EC2 AWS: Cloudformation

aboutte:
I took the [CloudFormation](https://github.com/trailofbits/algo/tree/CloudFormation) branch and fixed up a couple of things to get CF working for the EC2 deployment.

This PR can close  #132.

### DIFF
--- a/roles/cloud-ec2/templates/stack.yml.j2
+++ b/roles/cloud-ec2/templates/stack.yml.j2
@@ -38,12 +38,6 @@ Resources:
           Value: Algo
       VpcId: !Ref VPC
 
-  # SubnetIPv6:
-  #   Type: AWS::EC2::SubnetCidrBlock
-  #   Properties:
-  #     Ipv6CidrBlock: !Ref VPCIPv6
-  #     SubnetId: !Ref Subnet
-
   VPCGatewayAttachment:
     Type: AWS::EC2::VPCGatewayAttachment
     Properties:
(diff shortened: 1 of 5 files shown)


gunph1ld:
Agree, I think we need to handle IPv6 before merge, I made some drafts in the branch, but not sure if it works. I will check tonight

gunph1ld:
I'm not sure, but seems the [AWS docs](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpc.html) are wrong, I can not get the Ipv6CidrBlocks value form `AWS::EC2::VPC`

```
Outputs:
  Ipv6Cidr:
    Value: !GetAtt VPC.Ipv6CidrBlocks
```

The errors is:
`Template format error: The Value field of every Outputs member must evaluate to a String.`

gunph1ld:
New things implemented. @aboutte could you rebase with [the initial branch](https://github.com/trailofbits/algo/tree/CloudFormation)?

gunph1ld:
Nice. Fixes #60 also

aboutte:
@gunph1ld this is tested and working now.  I had to make a few changes to get ipv6 working correctly.  Because we have ipv6 forwarding turned on I had to add `sysctl net.ipv6.conf.\$IFACE.accept_ra=2`, otherwise I was not getting an ipv6 default route.  

screen shot 2017-04-18 at 10 17 25 am

dguido:
Awesome!

roles/cloud-ec2/templates/stack.yml.j2
```diff
@@ -0,0 +1,192 @@
...
+      - InstanceSecurityGroup
+    Metadata:
+      AWS::CloudFormation::Init:
+        config:
+          users:
+            ubuntu:
+              groups:
+                - "sudo"
+              homeDir: "/home/ubuntu/"
+          files:
```
defunctio: There is nothing wrong with this method, but I'm just wondering if there was a specific reason for this vs using ssh-authorized-keys param?

PR #433, opened 2017-04-17
Use dummy1 instead of lo:100 on Linux

kargig:
A dummy interface is better than using loopback for routing packets to/from.

### DIFF
--- a/roles/common/handlers/main.yml
+++ b/roles/common/handlers/main.yml
@@ -15,10 +15,13 @@
     ifconfig lo100 inet6 FCAA::1/64; echo $?
 
 - name: load dummy
-  shell: modprobe dummy
+  modprobe:
+    name: dummy
+    state: present
+    params: 'numdummies=2'
 
 - name: add dummy
-  shell: ip link add dummy100 type dummy
+  shell: ip link add dummy1 type dummy || /bin/true
 
 - name: restart dummy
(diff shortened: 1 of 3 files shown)


dguido:
@kargig Can you address the tests that are failing?

roles/common/handlers/main.yml
```diff
@@ -7,12 +7,18 @@
...
-
 - name: restart loopback bsd
   shell: >
     ifconfig lo100 destroy || true &&
     ifconfig lo100 create &&
     ifconfig lo100 inet {{ local_service_ip }} netmask 255.255.255.255 &&
     ifconfig lo100 inet6 FCAA::1/64; echo $?
+
+- name: load dummy
+  shell: modprobe dummy
```
gunph1ld: http://docs.ansible.com/ansible/modprobe_module.html

roles/common/handlers/main.yml
```diff
@@ -7,12 +7,18 @@
...
+
+- name: add dummy
+  shell: ip link add dummy100 type dummy
```
gunph1ld: Need to consider, that dummy will not be UP after reboot

kargig:
I am not sure if I can do anything about travis not being able to load the dummy module since it's run on an LXC container and afaik LXCs can't load kernel modules.

gunph1ld:
Closing due to inactivity

Issue #400, opened 2017-04-14
~/.ssh/config symlink overwritten


aboutte:
### OS / Environment
MacOS 10.11.5

### Ansible version
```
$ ansible --version
ansible 2.2.2.0
  config file =
  configured module search path = Default w/o overrides
```

### Version of components from `requirements.txt`
```
$ pip show msrestazure
Name: msrestazure
Version: 0.4.7
Summary: AutoRest swagger generator Python client runtime. Azure-specific module.
Home-page: https://github.com/Azure/msrestazure-for-python
Author: Microsoft Corporation
Author-email: UNKNOWN
License: MIT License
Location: /Users/aboutte/Documents/r/algo/env/lib/python2.7/site-packages
Requires: msrest, adal, keyring
(env)
aboutte@: ~/r/algo on master [!]
$
(env)
aboutte@: ~/r/algo on master [!]
$ pip show setuptools
Name: setuptools
Version: 34.4.1
Summary: Easily download, build, install, upgrade, and uninstall Python packages
Home-page: https://github.com/pypa/setuptools
Author: Python Packaging Authority
Author-email: distutils-sig@python.org
License: UNKNOWN
Location: /Users/aboutte/Documents/r/algo/env/lib/python2.7/site-packages
Requires: six, packaging, appdirs
(env)
aboutte@: ~/r/algo on master [!]
$ pip show ansible
Name: ansible
Version: 2.2.0.0
Summary: Radically simple IT automation
Home-page: http://ansible.com/
Author: Ansible, Inc.
Author-email: info@ansible.com
License: GPLv3
Location: /Users/aboutte/Documents/r/algo/env/lib/python2.7/site-packages
Requires: pycrypto, paramiko, setuptools, jinja2, PyYAML
(env)
aboutte@: ~/r/algo on master [!]
$ pip show dopy
Name: dopy
Version: 0.3.5
Summary: Python client for the Digital Ocean API
Home-page: https://github.com/devo-ps/dopy
Author: Vincent Viallet
Author-email: vincent@devo.ps
License: The MIT License (MIT)
Location: /Users/aboutte/Documents/r/algo/env/lib/python2.7/site-packages
Requires: requests
(env)
aboutte@: ~/r/algo on master [!]
$ pip show boto
Name: boto
Version: 2.46.1
Summary: Amazon Web Services Library
Home-page: https://github.com/boto/boto/
Author: Mitch Garnaat
Author-email: mitch@garnaat.com
License: MIT
[...]

PR #435, opened 2017-04-18
Removing update to ~/.ssh/config #400

aboutte:


dguido:
Nice, thanks! This looks good to me. Anyone else want to weigh in?

Issue #430, opened 2017-04-17
IKEv2 security association continually re-established when using iOS 10, strongSwan 5.3.5


evanchaney:
### OS / Environment

- iPhone SE, iOS 10.3.1 (14E304)
- strongSwan 5.3.5
- Ubuntu server 16.04.2 LTS
- Digital Ocean (droplet created by Algo @ 27e0fd0)

### Ansible version

- ansible 2.2.0.0

### Version of components from `requirements.txt`

- setuptools 34.3.2
- ansible 2.2.0.0
- dopy 0.3.5
- boto 2.46.1
- boto3 1.4.4
- azure 2.0.0rc5
- msrest 0.4.1
- apache-libcloud 1.5.0
- six 1.10.0
- pyopenssl 16.2.0
- jinja2 2.8

### Summary of the problem

strongSwan 5.3.5 when used with iOS 10 results in continual
re-establishment of the IKEv2 security association between the iOS
device and charon daemon. Re-association happens anywhere from
several times per minute to several times per hour.

When used on an everyday device with many services setup (email,
calendaring, background app refreshing) an apparent effect is
significantly decreased battery life on the iOS device (2x difference
observed between the IKEv2 VPN configuration and a non-IKEv2 VPN
configuration). No effect has been observed on the ability to retrieve
data over the VPN, however.

In one instance 51 re-associations were observed over 18 minutes.

### Steps to reproduce the behavior

1.  Install strongSwan 5.3.5 on Ubuntu 16.04 LTS using Algo.
2.  Enable charon logging to /var/log/charon.log (set `append=yes` in
    strongswan.conf, adjust AppArmor profile to allow access to
    /var/log/charon.log, reload AppArmor profile and restart
    strongswan).
3.  Perform restore of iOS 10.3.1 onto an iOS device. Set up device as
    new device (no restore from backup).
4.  Attach device to a Wi-Fi network.
5.  Install Algo IKEv2 VPN configuration profile.
6.  Let device go to sleep and remain in sleep.
7.  Observe the IKEv2 server sends the device dead-peer-detection
    notifications and receives no response. After several retries the
    IKEv2 server terminates the IKE security association requiring the
    iOS device to re-establish it and the child SA. Observe the issue
    repeats continuously.

### The way of deployment (cloud or local)

- Droplet on Digital Ocean

### Expected behavior

-   iOS device responds to dead-peer-detection request from IKEv2
    server. IKEv2 server retains the IKE SA between the server and
    client.

[...]

PR #437, opened 2017-04-18
disable dpd #430

dguido:
There's a subtle bug in strongSwan 5.3.5 that causes iOS clients to excessively re-associate to the server. This causes unnecessary battery drain, among other things. This can be avoided by disabling dead peer detection, or by upgrading to strongSwan 5.5.1. We chose to disable dpd rather than migrate our entire base install from 16.04 to 17.04.

More information is available in https://github.com/trailofbits/algo/issues/430


jauderho:
This is also probably why I'm in favor of a custom strongSwan build at some point.

PR #447, opened 2017-04-19
Updating documentation to include minimum required IAM polcy. Closes

### LINKED ISSUE #123: EC2: IAM inline policy

aboutte:
This PR is to close #123 

It is related to #364.  As soon as #431 is merged #364 will be out of date.


dguido:
I merged this elsewhere

PR #448, opened 2017-04-19
Change uniqueids back to never

dguido:
We need this to allow multiple connections with the same id/certificate


dguido:
We modified uniqueids when we removed dead peer detection. We had the best of intentions: without dead peer detection, we thought clients might start ghosting and fail to get flushed out. If we set uniqueids to "replace", then we could new connections with the same ID would replace old ones and prevent runaway ghost connections. But multiple connections per ID is an important use case we want to preserve so I think we should add this back.

Issue #449, opened 2017-04-19
Upgrade base image to Ubuntu 17.04


dguido:
Mainly for strongSwan 5.5.1. We care about 5.5.1 because it avoids the dead peer detection bug for macOS and iOS.

17.04 is available on EC2 and Digital Ocean already. It is not yet available on Azure or Cloud Engine yet.

dguido:
@gunph1ld Can you add a config flag to the config.cfg that lets you select 17.04? I think 17.04 will "just work" as long as the cloud hosting provider supports it (AWS and DigitalOcean do). Leave the default at 16.04 obviously.

davidemyers:
In 17.04 with the security role deployed I'm seeing some new log spam every time some bot tries to log in:

`Apr 23 12:44:38 vpn1 sshd[2607]: rexec line 28: Deprecated option RhostsRSAAuthentication`

Looks like SSH has changed from 16.04. This option can probably just be removed altogether since the `sshd_config` man page from 16.04 says,

     RhostsRSAAuthentication
             Specifies whether rhosts or /etc/hosts.equiv authentication
             together with successful RSA host authentication is allowed.  The
             default is “no”.  This option applies to protocol version 1 only.

And `sshd_config` already specifies only to use version 2.

bjornlll:
Any tentative date where we'll be able to run on Ubuntu 17.04? Alternatively, is there any way I can try out running on 17.04 right now? The VPN keeps dropping out on my iPhone, that's why I'm asking.

dguido:
16.04 works fine with our workaround.

As described in #466, distro version is configurable in config.cfg now.

gunph1ld:
All the cloud providers we use provide 17.04 now. We can switch to it permanently

bjornlll:
Curious: Is there any upside to switching now that 16.04 has the stongSwan patch we were waiting for?

PR #466, opened 2017-04-21
Let users choose the distro version

gunph1ld:
DO, AWS, GCE have ubuntu 17, but Azure still not

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ before_cache:
 
 env:
   - LXC_NAME=ubuntu1604 LXC_DISTRO=ubuntu LXC_RELEASE=xenial
+  - LXC_NAME=ubuntu1704 LXC_DISTRO=ubuntu LXC_RELEASE=zesty
 
 install:
   - sudo tar xf $HOME/lxc/cache.tar -C / || echo "Didn't extract cache."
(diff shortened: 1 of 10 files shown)


gunph1ld:
Fixes #449

dguido:
FYI the DNS Ad blocking breaks on 17.04 because of systemd haha

```
RUNNING HANDLER [dns_adblocking : restart dnsmasq] *****************************
fatal: [104.236.200.178]: FAILED! => {"changed": false, "failed": true, "msg": "Unable to restart service dnsmasq: Warning: dnsmasq.service changed on disk. Run 'systemctl daemon-reload' to reload units.\nJob for dnsmasq.service failed because the control process exited with error code.\nSee \"systemctl status dnsmasq.service\" and \"journalctl -xe\" for details.\n"}
```

jplock:
To fix it, I had to remove `/etc/dnsmasq.d/lxd` it was a symlink to no where for me

dguido:
Is there a patch we should make to that role so it works on both 16.04 and 17.04?

dguido:
Other than the adblocking role, it looks like all the other roles work perfect.

gunph1ld:
I've fixed the apparmor profile

dguido:
Nice! This is going to make it really easy to fully migrate to 17.04 when we want to.

Issue #307, opened 2017-03-29
Ubiquiti EdgeMax configuration


dmwyatt:
Anyone have any ideas or advice on how to configure an EdgeRouter Lite to send all internet traffic on LAN through the algo server?

Specifically, I'm looking for step-by-step instructions to enable my EdgeRouter Lite to route all internet traffic on my LAN through my algo VPN instead of the open internet.

Here's [my thread](http://w4t.pw/2y) on the ubiquiti forums linking to this issue.

[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=43597116)](https://www.bountysource.com/issues/43597116-ubiquiti-edgemax-configuration?utm_source=43597116&utm_medium=shield&utm_campaign=ISSUE_BADGE)

chriseldredge:
I just started looking at this too and found these community discussions that may be helpful:

* [EdgeRouter strongSwan configuration](https://community.ubnt.com/t5/EdgeMAX/vpn-routing-trouble-strongswan/td-p/866824)
* [Route all traffic through VPN](https://community.ubnt.com/t5/EdgeMAX/Route-all-traffic-through-VPN-client-pptpc0-EdgeMax-Pro/td-p/858298)

ebcodes:
Also very interested in this. 

On a side note, I wasn't even entirely sure this router would support Algo. The only thing I wasn't able to confirm after talking with Ubiquiti's support was support for ECDSA certificate keys. I did, however, find this in the release notes for version 1.8.0 of EdgeOS (it's currently on 1.9.1):

> [**IPsec**] Add the include-ipsec-secrets option for including a custom secrets file. This can be useful for example if an _ECDSA_ key is used. Implemented by TriJetScud

This tells me it can support it with the latest versions of EdgeOS, but once I get mine I'll test it out.

dmwyatt:
I found [this](https://help.ubnt.com/hc/en-us/articles/216771078-EdgeRouter-IPSec-VPN-CLI-Commands) which I _think_ is what we need, but I'm far from a networking guy so I'm not sure.

ebcodes:
```
root@ubnt:/etc/ipsec.d/certs# ipsec up ikev2-<ALGO_IP>
initiating IKE_SA ikev2-<ALGO_IP>[8] to <ALGO_IP>
generating IKE_SA_INIT request 0 [ SA KE No N(NATD_S_IP) N(NATD_D_IP) ]
[...]

PR #473, opened 2017-04-23
Adding support for EdgeMax clients

kiratp:
Known issues
1) IPV6 is untested, probably broken
2) multiple clients connecting to the server seems to break routing.

### DIFF
--- a/roles/vpn/templates/client_ipsec.conf.j2
+++ b/roles/vpn/templates/client_ipsec.conf.j2
@@ -28,7 +28,7 @@ conn ikev2-{{ IP_subject_alt_name }}
     leftsubnet={{ edgemax_lan_subnet }}
     auto=route
 {% else %}
-    audto=add
+    auto=add
 {% endif %}
 	
 	
--- a/roles/vpn/templates/client_ipsec.conf.j2
+++ b/roles/vpn/templates/client_ipsec.conf.j2
@@ -24,5 +24,18 @@ conn ikev2-{{ IP_subject_alt_name }}
     leftcert={{ item }}.crt
     leftfirewall=yes
     left=%defaultroute
-
-    auto=add
+{% if edgemax_support is defined and edgemax_lan_subnet is defined and edgemax_support == True %}
+    leftsubnet={{ edgemax_lan_subnet }}
+    auto=route
+{% else %}
+    audto=add
+{% endif %}
+	
+	
+{% if edgemax_support is defined and edgemax_lan_subnet is defined and edgemax_support == True %}
+conn lanbypass
+    leftsubnet={{ edgemax_lan_subnet }}
+    rightsubnet={{ edgemax_lan_subnet }}
+    type=passthrough
+    auto=route
+{% endif %}
--- a/config.cfg
+++ b/config.cfg
@@ -41,8 +41,8 @@ BetweenClients_DROP: Y
 # ALPHA feature: Uncomment below to build EdgeMaxclient configuration. Note that this will break connectivity from other clients
 # We recommend that you deploy a sepearate Algo server for each Edgerouter with only one client user defined.
 # Set the lan subnet to the IP CIDR notation for the lan behind the router. This MUST not interest with the vpn_network set above
-edgemax_support: True
-edgemax_lan_subnet: 10.0.0.0/16
+#edgemax_support: True
+#edgemax_lan_subnet: 10.0.0.0/16
 
 congrats:
   common: |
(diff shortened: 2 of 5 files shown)


ColdIce1:
Is it possible to update documenation to only mask a single IP from LAN-subnet through Algo IPSEC?

kiratp:
You should just connect directly from the client then. The router solution is for when you want to put most of your traffic through the tunnel.

roles/vpn/templates/client_ipsec.conf.j2
```diff
@@ -24,5 +24,18 @@ conn ikev2-{{ IP_subject_alt_name }}
     leftcert={{ item }}.crt
     leftfirewall=yes
     left=%defaultroute
-
-    auto=add
+{% if edgemax_support is defined and edgemax_lan_subnet is defined and edgemax_support == True %}
+    leftsubnet={{ edgemax_lan_subnet }}
+    auto=route
+{% else %}
+    audto=add
```
bjornlll: Spelling

dmwyatt:
What's the blockers on this PR?

gunph1ld:
I'm not a fan of such changes (not sure about the others) @dguido 
If we are going to support various routers we need to make it more flexible, but not just if-else statements

defunctio:
I agree with @gunph1ld on this one but for different reasons. The more features we accept from third parties ultimately results in the core team needing to support them in the future. I would also still really like to see a stable release cut.

On another note, I have been working on a refactor and CLI that supports and promotes the creation of third party roles for both providers and feature support. These providers and features can just be installed /upgraded/removed from the CLI as necessary with automatic module dependency resolution and can be maintained by the third party developers. Unfortunately it is not ready for testing at this time.

dmwyatt:
It seems like the ideal long-term solution would be a plugin-type of architecture.  What would be cool and fancy is if you could call algo with a git repo argument that pulls in the plugin automatically.  

On the one hand, there's security implications to making it easy to pull in third-party plugins.  On the other hand, I just don't see how it's possible for core devs to support everything that would be good to support.

Short-term is the idea that algo should just NOT support anything other than what the core dev team wants to support?  Maybe a compromise solution would be a section on a wiki detailing the changes needed for EdgeMax or whatever client or router?

Selfishly I'd like to see this PR merged (assuming it works)!  The poor guy worked hard on it!

defunctio:
Short-term isn't that we only accept things that we want to support. It's just that I'd love to see a release version cut before introducing new features but I've been aiming for that for awhile. We really need a stable branch so that people interested in using 'bleeding-edge' features can without effecting the rest of the userbase. Ultimately I'd like to see all features refactored as independent roles that can be included optionally.

I'm just one of the developers here but these goals are based on an alpha quality private fork at this stage it is unclear whether or not this concept will become part of Algo in the future. But the project goals are exactly that. In regards to the security implications, there will be an authorized list of 'trusted' developer teams that are only allowed be installed by default. This can be overridden by the end user after accepting a disclosure. These will be implemented through ansible-galaxy w/ additional metadata for Algo.

The way i see it this way we can focus on providing a core platform for the community to contribute to via role plugins that does not introduce breaking changes, maintenance and support issues into mainline Algo. I'll probably break this up into something like core, contrib and untrusted.

ndfred:
@kiratp: you don't mention hardware offload support here though you do mention it in #307 and your change discourages HW offload because it is unreliable. You also mention that routing might radomly break as other clients connect to the VPN server. Should these issues be addressed before adding official Algo support for EdgeMax? Are there any related discussions on the [community forums](https://community.ubnt.com/t5/EdgeMAX/bd-p/EdgeMAX) to explore these issues?

It is fantastic to see this working at all and I cannot wait to have it on my ERL (and you deserve full credit for getting us that far already!). I wouldn't use it until it is reasonably stable and performant though, and I would assume the same from anyone using Algo VPN.

dguido:
@ndfred: We're not adding official support for EdgeMax devices to Algo. We're refactoring the project into smaller, more modular parts. When that work is complete, you can probably expect to find a module for EdgeMax support. For now, you're on your own if this is the route you want to go.

kiratp:
@ndfred - UBNt thread regarding offload - https://community.ubnt.com/t5/EdgeMAX/IPSec-performance-issue/m-p/1946992#M162999

I plan to redo this PR once the refactor is complete.

ndfred:
@dguido: how far do you think we are from a stable release that would unlock the modularity work (the project's readme doesn't share where we are in the release process)? How bif a deal is Algo modularity vs other features?

@kiratp: super thorough research again on the offload bugs on that thread, would be cool to see Ubiquiti engineers pick it up. About the routing problems you are mentioning in this PR's description, do you have an idea of wether that would be on the ERL or Algo side of things? (assuming Algo since it depends on the number of VPN clients)

PR #502, opened 2017-04-29
break up "step 3" into two actions

neuroradiology:
old steps:
  3. install dependencies (but first cd to algo-master)

new steps:
  3. cd to algo-master
  4. install dependencies...


dguido:
Thanks, but I think it's fine as it is.

PR #510, opened 2017-05-02
Enhance PS1 script

### LINKED ISSUE #454: Windows: Ability to undo/reverse everything
For now, that is a relatively easy way to add VPN. But to remove VPN plus all the CA Cert is bit more complicated. (At least for me)

So I wonder if we should add a powershell script that does exactly the opposite of the currently-implemented ps1.

P.S. Is there a easier way to run ps1 without run command first.

gunph1ld:
Fixes #454

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,31 +102,16 @@ No version of Android supports IKEv2. Install the [strongSwan VPN Client for And
 
 ### Windows
 
-Windows clients have a more complicated setup than most others. Follow the steps below to set one up:
+#### Scripted installation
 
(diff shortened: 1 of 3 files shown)


dguido:
Neat! Needs docs.

tony199555:
Looks great. Thank you for your work.

tony199555:
I will try do some test today, but I do need some docs for more specific actions too. If you could provide some instruction that would be great.

gunph1ld:
> Do I still need to add CA before running this script?

No, you don't

> Do you see we can do something about it, too?

Just run the script as Administrator, it works for me and adds the CA itself

gunph1ld:
added some new docs

PR #515, opened 2017-05-03
Get strongswan from the Zesty repo on Xenial

### LINKED ISSUE #430: IKEv2 security association continually re-established when using iOS 10, strongSwan 5.3.5

gunph1ld:
I would prefer this trick instead of using ppa.
Fixes #430

### DIFF
--- a/roles/vpn/tasks/ubuntu-hacks.yml
+++ b/roles/vpn/tasks/ubuntu-hacks.yml
@@ -1,0 +1,14 @@
+---
+
+- name: Configure apt to use the Xenial release by default
+  template: src=01_xenial_aptconf.j2 dest=/etc/apt/apt.conf.d/01xenial
+
+- name: Configure packages preferences
+  template: src=01_strongswan.pref.j2 dest=/etc/apt/preferences.d/01_strongswan.pref
+
+- name: Configure the Ubuntu Zesty repository
+  apt_repository:
+    repo: deb http://mirrors.kernel.org/ubuntu/ zesty main
+    state: present
+    filename: 'zesty'
+    update_cache: yes
--- a/roles/vpn/tasks/ubuntu.yml
+++ b/roles/vpn/tasks/ubuntu.yml
@@ -3,6 +3,9 @@
 - set_fact:
     strongswan_additional_plugins: []
 
+- include: ubuntu-hacks.yml
+  when: ansible_distribution_version == "16.04"
+
 - name: Ubuntu | Install strongSwan
   apt: name=strongswan state=latest update_cache=yes install_recommends=yes
 
--- a/roles/vpn/templates/01_strongswan.pref.j2
+++ b/roles/vpn/templates/01_strongswan.pref.j2
@@ -1,0 +1,3 @@
+Package: *strongswan*
+Pin: release n=zesty
+Pin-Priority: 9000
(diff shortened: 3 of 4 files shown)


defunctio:
Have you tested this? I like this idea much better than a PPA. I don't think we should ever rely on a 3rd party PPA.

gunph1ld:
Yes, it worked

PR #512, opened 2017-05-02
Move to Elastic IP

gunph1ld:
Fixes #394

PR #521, opened 2017-05-06
Specify EIP domain

cjp:
CloudFormation throws the error `You must specify an allocation id when mapping an address to a VPC instance` when creating the ElasticIP.

This specifies the EIP's domain as `vpc`, and makes it dependent on the `VPCGatewayAttachment`, per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-eip.html.


gunph1ld:
Does it fix https://github.com/trailofbits/algo/pull/512#issuecomment-299515112 ?

PR #527, opened 2017-05-07
Add additional delay for ec2 instance prior to ssh

cjp:
Intermittently (in my experience, 5 out of 6 tries), algo attempts to connect to the ec2 instance before the public key is added by CloudFormation.

The following error is thrown:
```fatal: [x.x.x.x]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: Warning: Permanently added 'x.x.x.x' (ECDSA) to the list of known hosts.\r\nPermission denied (publickey).\r\n", "unreachable": true}```

Note, this does not appear to be related to the ssh directory permission error (https://github.com/trailofbits/algo/issues/321).

As indicated by https://github.com/trailofbits/algo/issues/429#issuecomment-295076148, a redeploy typically solves this.

By adding an additional 20 second delay prior to executing the `uname -a` in `playbooks/post.yml`, this error has ceased for me.

**I admit that this is an inelegant solution.** I've been unsuccessful in getting the `raw:` command to continue when an unreachable state is triggered.

### DIFF
--- a/playbooks/post.yml
+++ b/playbooks/post.yml
@@ -12,6 +12,6 @@
 
 - name: A short pause, in order to be sure the instance is ready
   pause:
-    seconds: 10
+    seconds: 20
 
 - include: local_ssh.yml
--- a/roles/cloud-ec2/tasks/main.yml
+++ b/roles/cloud-ec2/tasks/main.yml
@@ -62,10 +62,6 @@
         line: "{{ item.public_ip_address }}"
       with_items:
         - "{{ algo_instances.instances }}"
-
-    - name: Extra wait time for ec2 instance to be ready
-      pause:
-        seconds: 20
   rescue:
     - debug: var=fail_hint
       tags: always
--- a/roles/cloud-ec2/tasks/main.yml
+++ b/roles/cloud-ec2/tasks/main.yml
@@ -62,6 +62,10 @@
         line: "{{ item.public_ip_address }}"
       with_items:
         - "{{ algo_instances.instances }}"
+
+    - name: Extra wait time for ec2 instance to be ready
+      pause:
+        seconds: 20
   rescue:
     - debug: var=fail_hint
       tags: always

### CODE REVIEW

**Comment by dguido** [comment]
What if we add 10 seconds to the pause already in post.yml?

https://github.com/trailofbits/algo/blob/master/playbooks/post.yml

**Comment by cjp** [comment]
That works equally well. I was concerned users of unaffected cloud providers would see this as an unnecessary delay.

**Comment by dguido** [comment]
Ehh it's only an extra 10s, and if it causes less people to have issues, then it's fine.

**Comment by kurtharriger** [comment]
Even with the 20 second delay I failed twice deploying to aws us-west-2.  
I was starting to think I would need to hack, but kicked it off once more while I started checking open issues and found this...  of course as soon as I find this I see my third deploy seems to be working successfully.   Perhaps there is a way to pause, try, pause, retry, a few times rather than arbitrarily increasing it... but 20 seconds still seems to be a bit short for some environments.

**Comment by shapiro125** [comment]
For DO, the 20 seconds is fine. But, for EC2, I've manually updated https://github.com/trailofbits/algo/blob/master/playbooks/post.yml to a minute or more and it works perfectly:

```- name: A short pause, in order to be sure the instance is ready
  pause:
seconds: 90```

Agreed, would be very irritating to wait that long in other networks, but this is a good short term solution.

**Comment by emkman** [comment]
FYI this is still an issue on EC2. Attempted to install from master and SSH failed.
```
PLAY [Configure the server and install required software] **********************

TASK [Check the system] ********************************************************
fatal: [34.236.48.176]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: Warning: Permanently added '34.236.48.176' (ECDSA) to the list of known hosts.\r\nPermission denied (publickey).\r\n", "unreachable": true}

PLAY RECAP *********************************************************************
34.236.48.176              : ok=0    changed=0    unreachable=1    failed=0
localhost                  : ok=19   changed=5    unreachable=0    failed=0
```
I don't see any pause or delay steps running. I never attempted any other cloud besides AWS so I don't believe the ECDSA issue mentioned in the FAQ is the issue.

PR #536, opened 2017-05-09
cleaner syntax for local actions

rvben:
Good evening! :)

The use of the 'local_action' module is often less clean, as it makes code look very different, while the only thing that you want different is the host on which it runs. It's therefore cleaner to use "delegate_to: localhost", which can also be used on a higher level (includes/blocks) to make it even cleaner.

What do you think?

### DIFF
--- a/playbooks/local.yml
+++ b/playbooks/local.yml
@@ -1,16 +1,23 @@
 ---
 
 - name: Generate the SSH private key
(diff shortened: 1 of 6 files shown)


gunph1ld:
Great, Thanks. Can you make all the openssl commands multiline?

rvben:
Sure, that definitely makes it better. I made the changes.

rvben:
Of course. It's now rebased. :)

rvben:
Added #544 to this PR.

PR #544, opened 2017-05-11
Removed unnecessary local_action's

rvben:
I found a number of uses of local_action, while they were running on localhost anyway. Cleaned them up.

Have a great day!


gunph1ld:
Move to #536, please

rvben:
Excuse me, I overlooked the fact that that one wasn't merged. It should be added there, indeed.

PR #548, opened 2017-05-13
Make DNS blocklist URLs configurable

rvagg:
pull blocklist urls into config.cfg, some additional minor refactoring of adblock.sh as well


dguido:
Looks good to me!

PR #573, opened 2017-05-30
Increase the default size of the SSH key.

lazzurs:
The default SSH key size in 2048. Increasing this to 4096 to match
current best practice guidance.


dguido:
2048 is more than sufficient, especially given the short lived nature of most Algo installs. We'll transition to ECC keys when all our providers support them.

PR #575, opened 2017-05-30
Make documentation on iptables for local installation clearer.

devries:
I was looking at the instructions for installing on an existing ubuntu server and found the instructions regarding skipping iptables configuration confusing. This is my draft at a fix. I'm not a very advanced ansible user, so I think the `--skip-tags iptables` needs to be explicit. If not, I can rework the documentation in another pull request.


dguido:
Thanks!

Issue #577, opened 2017-05-31
Is GCE still only for testing?


lperry:
Sorry if this has been answered already (I checked the FAQ, troubleshooting docs, and existing issues), but is GCE still only for testing? On [line 412 in algo](https://github.com/trailofbits/algo/blob/master/algo#L412), GCE is called out for testing only, but the issue it references (https://github.com/trailofbits/algo/issues/369) has already been closed (and looks resolved).

Thanks!

dguido:
It should work now

Issue #369, opened 2017-04-08
Intermittent routing failures with GCE servers


dguido:
Algo (e.g., strongSwan) servers on Google Cloud Engine intermittently fail to route packets from VPN clients. In general, this manifests as clients being able to connect to the VPN but then unable to browse any or some websites through it. For example, the entire internet works but google.com won't.

We have been under the impression that MSS/MTU issues were to blame but adjusting these values does not appear to resolve the problem for most users. It's been reported more frequently than any other issue. It is likely that issues #358, #345, #310, #210, #185 are all related.

References we've found that may be helpful:
* [Enable IP forwarding for instances](https://cloud.google.com/compute/docs/networking#canipforward)
* [Maximum Transfer Unit (MTU) considerations](https://cloud.google.com/compute/docs/vpn/advanced#maximum_transfer_unit_mtu_considerations)

Please add any specific troubleshooting steps you've tried that did or did not work to this issue so we can keep track of them. Thanks!

kiratp:
(Disclosure - Google employee, posting in my own capacity)

More info:

Something in the Ubuntu image is not not interacting with the user lying infrastructure. I managed to get the tunnel fully working by

1. Launch a VM using the Debian image and make sure IP forwarding is enabled
2. SSH in and install git + the algo dependancies. 
3. Skip the virtualenv setup and install the python dependancies directly (```sudo pip install -r requirements.txt```). You'll have to fight your way through some stuff here - rerunning the install gets you working but it will complain about Azure dependency mismatches.
4. Copy keys to client and connect. The tunnel should work but the MTU will cause website load failures.
5. ```sudo iptables -t mangle -A FORWARD -o eth0  -p tcp -m tcp --tcp-flags SYN,RST SYN   -j TCPMSS --set-mss 1280```

Tunnel works fully - using one as I type this.

However, if I perform a local install (with virtualenv) on the ubuntu image, the tunnel does not work. The client can connect and pings to 8.8.8.8 seem to work but DNS requests are black-holed - dig google.com just hangs.

I looked at tcpdumps from the servers. On Ubuntu, I can see that the server forwards the DNS request, and I can see the response ESP packet being sent but it never arrives at the client. 

Next steps are to try and figure out how the response packets are different between Ubuntu and Debian. Its going to be difficult given version differences 

Output from ```ipsec version```
Debian - Linux strongSwan U5.2.1/K3.16.0-4-amd64 
[...]

PR #618, opened 2017-07-08
Add additional GCE zones (London / Australia)

cdolle:
Adds support for GCE zones in London (europe-west2 a-c) and Australia (australia-southeast1 a-c)

Removes "For testing only" text when setting up GCE (#369) - this issue seems completely fixed now and removing the warning will eliminate future issues like #577


dguido:
Thanks!

PR #631, opened 2017-07-22
Fix for Hash Sum mismatch during Ubuntu | Install prerequisites

Dhakchianandan:
apt-get update fails due to hash sum mismatch. This stops ./algo installation going forward.

Fix based on: 
[](https://askubuntu.com/questions/41605/trouble-downloading-packages-list-due-to-a-hash-sum-mismatch-error)


dguido:
Thanks @ghost! Can you rebase your changes?

Issue #436, opened 2017-04-18
Unable to reach websites from OpenSUSE Linux client


hirozed:
### OS / Environment
Ubuntu 16.04 VPS on OVH hosting

### Ansible version
ansible 2.2.0.0

### Version of components from `requirements.txt`
Name: msrestazure
Version: 0.4.7
Name: setuptools
Version: 35.0.0
Name: ansible
Version: 2.2.0.0
Name: dopy
Version: 0.3.5
Name: boto
Version: 2.46.1
Name: boto3
Version: 1.4.4
Name: msrest
Version: 0.4.1
Name: apache-libcloud
Version: 1.5.0
Name: six
Version: 1.10.0
Name: pyOpenSSL
Version: 16.2.0
Name: Jinja2
Version: 2.8

### Summary of the problem
I have successfully set up Algo on an Ubuntu 16.04 server.  My iOS & OS X devices connect and can access the the internet without issue.  

Connecting via my Linux box (OpenSUSE Tumbleweed) through Network Manager fails.  I have installed Algo with Windows 10 settings, introducing the weaker encryption.  I'm honestly not sure if I am using the correct certificates, but I tried different ones.

Going via strongswan directly, connects to the server, but I don't seem to have access to DNS.  I can ping ip addresses, but not domain names.  

Since I have had better success via command line strongswan, that might be a better route to go.  I can provide any other information you need.

### Steps to reproduce the behavior
Set up a linux box and try connecting via Network Manager or strongswan cli.

### The way of deployment (cloud or local)
Local

### Expected behavior
Connect to VPN and successfully ping a domain.

### Actual behavior
Pinging returns nothing, not even an error message.  Pinging via IP goes through

### Full log
The installation itself seems sound, I unfortunately do not have the log.

Any help you can provide would be fantastic.

-Jim

dguido:
As far as I can tell, OpenSuse Tumbleweed uses Network Manager 1.4.1 (https://software.opensuse.org/package/NetworkManager-strongswan) which is the same as every other current Linux distribution (Debian Stretch, Fedora 25, Ubuntu 17.04, etc). We have many reports of these Linux distributions working fine as long as you're not using AES-GCM (which Network Manager 1.4.1 does not support). I don't think this is a bug with Algo. You may want to join our Slack for helping troubleshooting this issue.

molecularswords:
[...]

Issue #356, opened 2017-04-05
Known issue with virtualenv and ansible on SELinux-enabled hosts


postmodern:
### OS / Environment
Fedora Linux 25

### Ansible version
2.2.1.0

### Version of components from `requirements.txt`
```
Name: msrestazure
Version: 0.4.7
Summary: AutoRest swagger generator Python client runtime. Azure-specific module.
Home-page: https://github.com/Azure/msrestazure-for-python
Author: Microsoft Corporation
Author-email: UNKNOWN
License: MIT License
Location: /home/postmodern/algo/env/lib/python2.7/site-packages
Requires: keyring, adal, msrest
---
Name: setuptools
Version: 34.3.3
Summary: Easily download, build, install, upgrade, and uninstall Python packages
Home-page: https://github.com/pypa/setuptools
Author: Python Packaging Authority
Author-email: distutils-sig@python.org
License: UNKNOWN
Location: /home/postmodern/algo/env/lib/python2.7/site-packages
Requires: appdirs, packaging, six
---
Name: ansible
Version: 2.2.0.0
Summary: Radically simple IT automation
Home-page: http://ansible.com/
Author: Ansible, Inc.
Author-email: info@ansible.com
License: GPLv3
Location: /home/postmodern/algo/env/lib/python2.7/site-packages
Requires: PyYAML, jinja2, paramiko, pycrypto, setuptools
---
Name: dopy
Version: 0.3.5
Summary: Python client for the Digital Ocean API
Home-page: https://github.com/devo-ps/dopy
Author: Vincent Viallet
Author-email: vincent@devo.ps
License: The MIT License (MIT)
Location: /home/postmodern/algo/env/lib/python2.7/site-packages
Requires: requests
---
Name: boto
Version: 2.46.1
Summary: Amazon Web Services Library
Home-page: https://github.com/boto/boto/
Author: Mitch Garnaat
Author-email: mitch@garnaat.com
License: MIT
Location: /home/postmodern/algo/env/lib/python2.7/site-packages
Requires: 
---
Name: boto3
Version: 1.4.4
Summary: The AWS SDK for Python
Home-page: https://github.com/boto/boto3
Author: Amazon Web Services
Author-email: UNKNOWN
License: Apache License 2.0
Location: /home/postmodern/algo/env/lib/python2.7/site-packages
Requires: jmespath, botocore, s3transfer
---
Name: azure
Version: 2.0.0rc5
Summary: Microsoft Azure Client Libraries for Python
[...]

Issue #300, opened 2017-03-29
Fedora client connects to server but no traffic passes through


molecularswords:
### OS / Environment
Fedora

### Ansible version

2.2.0.0

### Version of components from `requirements.txt`

setuptools 34.3.3
dopy 0.3.5
boto 2.46.1
boto3 1.4.4
azure 2.0.0rc5
msrest 0.4.1
apache-libcloud 1.5.0
six 1.10.0
pyopenssl 16.2.0
jinja2 2.8

### Summary of the problem
After deploying a client setup using the deploy_client.yml I can successfully make the connection to the VPN server (on a DO droplet) using the command "sudo strongswan up <my connection name here>" but no traffic goes through the server. The server is working fine as a VPN as assessed from my Android client, but nothing will resolve using the Fedora client.

### Steps to reproduce the behavior
1) Deploy the client with parameters client_ip=localhost, vpn_user=<the username I chose during setup>, server_ip=<the server's ip>, server_ssh_user=root.
2) Make the connection with the command "sudo strongswan up <my connection name here>" and receive confirmation that the connection was established.
3) Ping any domain, go to any website in a browser, or open a program that requires a working internet connection and nothing happens due to a lack of traffic getting through.

### The way of deployment (cloud or local)
cloud (DO droplet)

### Expected behavior
Given that the connection is successfully established I expect the traffic to properly route through the server.

### Actual behavior
The traffic does not properly route through the server.

molecularswords:
I should mention that I tested the MTU/MSS fix that some people are experiencing, and it did not solve the problem.

dguido:
There are more precise instructions for strongSwan clients on the readme. You should try them. What you're doing is generally unsupported by our project.

molecularswords:
I have followed the instructions from the beginning without success on Fedora but with success on an Ubuntu VM. The only difference between what one has to do on Fedora and what the readme says is to use the "strongswan" command instead of the "ipsec" command because the former replaces the latter on Fedora. I also had to alter the location of the config files and certs since on Fedora the ipsec.d directory is under /etc/strongswan/ and not just /etc/. It works fine in an Ubuntu VM, and all the commands for strongswan and ipsec are the same. I'm not sure why what I'm doing is not supported since what I'm doing is essentially no different than the readme instructions. It seems to just be a difference between Fedora and Ubuntu that I have not yet discovered how to fix. If anyone has gotten it working on a Fedora client please give us some insight.

PR #711, opened 2017-10-30
Docs to deploy from, and setup client on, Fedora Workstation

sjugge:
This PR introduces docs to:

* deploy Algo from Fedora Workstation
* install and use a client on Fedora Workstation
* reference the above appropriatly

There are some challenges in order to install / deploy / use Algo on Fedora and information is scattered in the docs and issues. Docs include info sourced from links below and have been tested on latest stable.

Sources aside from existing docs (also referencing these in case it may help someone out):

* #263 This issue especially was a treasure trove of insights and patched goodness
* #356
* #436 
* #300 
* #372

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,7 @@ After this process completes, the Algo VPN server will contains only the users l
 
 * Setup instructions
   - Documentation for available [Ansible roles](docs/setup-roles.md)
+  - Deploy from [Fedora Workstation (26)](docs/deploy-from-fedora-workstation.md)
   - Deploy from [RedHat/CentOS 6.x](docs/deploy-from-redhat-centos6.md)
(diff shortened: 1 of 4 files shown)


dguido:
Nice! I'll review this within a day or two.

dguido:
Sorry for the wait, I spent all last week at 3 different conferences. I'm going to accept your PR and we can fix any issues that pop up in future commits. It's difficult for me to test this beyond simply reviewing it for English grammar since I don't use Fedora, so we'll have to let the community hash it out.

sjugge:
Sounds good. Fedora Workstation 27 just came out this week, from the looks of it a few things may be behaving wonky after the upgrade. I need to dig a bit deeper into this but I'll report back if need be through a PR or issue.

Issue #735, opened 2017-11-21
adblock.sh cron job fails restart dnsmasq service


melizeche:
### OS / Environment
Ubuntu 16.04.3 LTS
cloned from master@e01521bbf493d04dd5ffc7942b558d50674b414f

### Ansible version
ansible 2.2.0.0

### Summary of the problem
adblock.sh cron job fails to restart dnsmasq service

### Steps to reproduce the behavior
./algo
Do you want to install a DNS resolver on this VPN server, to block ads while surfing?: y

### The way of deployment (cloud or local)
cloud

### Expected behavior
cron job to finish succesfully

### Actual behavior
root got mail from crontab saying "service: not found"
```
Date: Mon, 20 Nov 2017 02:10:20 GMT
From: Cron Daemon <root@xxx>
To: root@xxx
Subject: Cron <root@xxx> /usr/local/sbin/adblock.sh

Downloading hosts lists...
/usr/local/sbin/adblock.sh: 41: /usr/local/sbin/adblock.sh: service: not found
```

melizeche:
adblock.sh currently uses `service dnsmasq restart`, that's the Upstart way to do it, algo runs on Ubuntu 16.04.3 that uses systemd so it should use `systemctl restart dnsmasq.service`

PR #736, opened 2017-11-21
Update adblock.sh for systemd to fix issue #735

melizeche:
Currently adblock.sh fails on line 41 when is called from the cron job because of the lack of PATH for `/usr/sbin/` that is where `service` is located. But `service dnsmasq restart` is the Upstart way to restart a service, since Ubuntu 16.04.3 use systemd the best option IMO is to use `systemctl restart dnsmasq.service`, `systemctl` command is located in `/bin` that currently is the PATH for adblock.sh when is called from the cron job, so no other modification is needed.

Also I updated the script following the recommendations of shellcheck to fix https://github.com/koalaman/shellcheck/wiki/SC2006
https://github.com/koalaman/shellcheck/wiki/SC2196

PS: The fix for #735 is in the first commit and the shellcheck improvements are in separated commits just in case.

### DIFF
--- a/roles/dns_adblocking/templates/adblock.sh.j2
+++ b/roles/dns_adblocking/templates/adblock.sh.j2
@@ -1,8 +1,8 @@
 #!/bin/sh
 # Block ads, malware, etc..
 
-TEMP=`mktemp`
-TEMP_SORTED=`mktemp`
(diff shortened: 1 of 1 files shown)


dguido:
Thank you!

PR #784, opened 2018-01-20
DigitalOcean - Add cleanup step for SSH key

douggard:
Currently, SSH keys are only deleted prior to uploading a new one. If you are using them with dynamic names though (configurable in config file), you will end up cluttering your DigitalOcean account with SSH keys that will never been delete. 

This is mostly just a copy paste of the delete step, but at the end of the cloud-digitalocean task to act as cleanup.


dguido:
👍

PR #804, opened 2018-02-16
Ansible 2.4, Lightsail, Scaleway, DreamCompute (OpenStack) integration

### LINKED ISSUE #808: Python msrestazure/msrest dependency mismatch

```
Name: msrestazure
Version: 0.4.21
Summary: AutoRest swagger generator Python client runtime. Azure-specific module.
Home-page: https://github.com/Azure/msrestazure-for-python
Author: Microsoft Corporation
Author-email: azpysdkhelp@microsoft.com
License: MIT License
Location: /private/tmp/algo/env/lib/python2.7/site-packages
Requires: msrest, keyring, adal
---
Name: msrest
Version: 0.4.1
Summary: AutoRest swagger generator Python client runtime.
Home-page: https://github.com/xingwu1/autorest/tree/python/ClientRuntimes/Python/msrest
Author: Microsoft Corporation
Author-email: UNKNOWN
License: MIT License
Location: /private/tmp/algo/env/lib/python2.7/site-packages
Requires: requests, chardet, certifi, requests-oauthlib, enum34, keyring, isodate
```

### Summary of the problem

The current `requirements.txt` pins `msrest` to version 0.4.1 but does not specify a version for the `msrestazure` package. Since no version is specified for `msrestazure` the newest version, 0.4.21 as of this writing, is installed. This version depends on `msrest>=0.4.25,<2.0.0` (see [setup.py](https://github.com/Azure/msrestazure-for-python/blob/v0.4.21/setup.py#L52)).

The installation of the requirements via `pip` *works* because top-level requirements take precedence over sub-dependencies requirements (<https://github.com/pypa/pip/issues/988>).

I do not use Azure, am not familiar with the python packages in question, and do not know why `msrest` has been pinned to version 0.4.1. Furthermore, I do not know if the dependency mismatch actually causes problems for Algo.

I ran into this problem when I tried to install the requirements via [pipenv](https://github.com/pypa/pipenv). 

```
algo$ pipenv install --two --requirements requirements.txt
[ ... ]
Warning: Your dependencies could not be resolved. You likely have a mismatch in your sub-dependencies.
  You can use $ pipenv install --skip-lock to bypass this mechanism, then run $ pipenv graph to inspect the situation.
Could not find a version that matches msrest<2.0.0,==0.4.1,>=0.4.25
Tried: 0.0.1, 0.0.1, 0.0.2, 0.0.2, 0.0.3, 0.0.3, 0.1.0, 0.1.0, 0.1.1, 0.1.1, 0.1.2, 0.1.2, 0.1.3, 0.1.3, 0.2.0, 0.2.0, 0.3.0, 0.3.0, 0.4.0rc1, 0.4.0rc1, 0.4.0, 0.4.0, 0.4.1, 0.4.1, 0.4.2, 0.4.2, 0.4.3, 0.4.3, 0.4.4, 0.4.4, 0.4.5, 0.4.5, 0.4.6, 0.4.6, 0.4.7, 0.4.7, 0.4.8, 0.4.8, 0.4.9, 0.4.9, 0.4.10, 0.4.10, 0.4.11, 0.4.11, 0.4.12, 0.4.12, 0.4.13, 0.4.13, 0.4.14, 0.4.14, 0.4.15, 0.4.15, 0.4.16, 0.4.16, 0.4.17, 0.4.17, 0.4.18, 0.4.18, 0.4.19, 0.4.19, 0.4.20, 0.4.20, 0.4.21, 0.4.21, 0.4.22, 0.4.22, 0.4.23, 0.4.23, 0.4.24, 0.4.24, 0.4.25, 0.4.25, 0.4.26, 0.4.26
[ ... ]
```

### Expected behavior

Python requirements can be installed without sub-dependencies mismatches.

### Actual behavior

Installation of python requirements fails iff sub-dependencies are taken into account and verified.

jackivanov:
**List of changes:**
- Ansible 2.4 compatibility
- Unattended-upgrades enabled by default for all new cloud instances
- Security role deleted
- Lightsail support
- Scaleway support (`uri` module)
- OpenStack support (DreamCompute optimized, but the role might be used to deploy to any OpenStack environment)
- Update requirements.txt. Fixes #808

Lightsail module introduced in Ansible 2.4, but doesn't support some important features and have a [bug](https://github.com/ansible/ansible/issues/36313). 
Let's keep a local module until [the changes](https://github.com/ansible/ansible/pull/36316) merged.

Need to test all the cloud provides before merge
- [x] DigitalOcean
- [x] Amazon Lightsail
- [x] Amazon EC2
- [x] Microsoft Azure
- [x] Google Compute Engine
- [x] Local


dguido:
Let’s remove the security role. It does very little, especially since we upgraded the base image to Ubuntu 17.10.

gunph1ld:
OK. The major update is coming. Added 3 more providers - Lightsail, Scaleway, DreadmCompute. 
DreamCompute is actually an openstack based provider, so anyone can use this role to deploy on any OpenStack environment.

ping @ndtob @joshuaiz @computerality

davidemyers:
I tested a Lightsail deployment. It works, but I have a few comments.

The instance is not assigned a static IP address so the address could conceivably change during a reboot. See [Static IP addresses in Amazon Lightsail](https://lightsail.aws.amazon.com/ls/docs/overview/article/understanding-static-ip-addresses-in-amazon-lightsail).

`dnsmasq` is picking up an extra server (172.26.0.2). Uncommenting `no-resolv` in `/etc/dnsmasq.conf` gets rid of it.

The Lightsail firewall has port 80 open (not a big deal, of course).

Thanks for making Lightsail an option!

dguido:
I just ran a deploy on macOS 10.13 to DigitalOcean and it all works as expected.

davidemyers:
I tested a deployment to Ubuntu 17.10 on DigitalOcean. It works fine except for an issue with `dnsmasq` like the one I mentioned above.

In addition to the DNS servers I specify in `config.cfg`, `dnsmasq` is reading `/etc/resolv.conf` and adding the server `127.0.0.53`, which belongs to `systemd-resolved`, which consults a different set of servers as specified by DigitalOcean in `/etc/netplan/50-cloud-init.yaml`. The configuration option `no-resolv` fixes this.

I should admit that I only use `dnsmasq` as a caching forwarder so that I can resolve local host names by putting them in `/etc/hosts`. Since I install ad blockers on each of my clients I disable the use of `dnsmasq` as an ad blocker post installation. But I really want only the DNS servers I specify in `config.cfg` to be used (Quad9 in my case).

jackivanov:
@davidemyers Thanks! I believe it should be covered in another issue

PR #809, opened 2018-02-19
Use default nameservers which given by the provider instead of googles

### LINKED ISSUE #806: DNS Leak(s) and Plugging/Preventing Them - Suggestions???
### OS / Environment

1. VPN is on Digital Ocean UBUNTU 16.04
2. Local Machine is Mac running macOS High Sierra Version 10.13.3
3. Running DNS Leaks on Safari and Firefox. 
4. WebRTC is turned off
5. General search is through DuckDuckGo, cache is cleared completely before session in browser is opened.(clean slate)

Running the following plugins on each browser

- [HTTPSEverywhere](https://www.eff.org/https-everywhere)
- [BuiltWith](https://builtwith.com/)
- [Momentum](https://momentumdash.com/)
- [Highfive](https://highfive.com/)

### Ansible version
`ansible - version 2.2.1`

### Summary of the problem

DNS is leaking

1. Spin up ALGO VPN on Digital Ocean
2. Connect to VPN from Local Machine
3. Run Standard DNS Leak Test from [https://www.dnsleaktest.com/](https://www.dnsleaktest.com/) and [https://whoer.net/](https://whoer.net/)
4. Confirm that IP addresses in leak tests are not the VPN address of the server running on Digital Ocean

Cloud

### Expected behavior
I expect the only IP address to show up to be the IP of the VPN server.  No DNS should be leaking.

### Actual behavior
The DNS Lists a bunch of Google IP addresses

Such as 
74.125.44.132 (which traces to Google)
74.125.19.12 (which traces to Google)
74.125.44.66 (which traces to Google)
74.125.19.2 (which traces to Google)
74.125.19.8 (which traces to Google)
74.125.44.74 (which traces to Google)

### Full log
No log needed.  This is a question about DNS leaks, not sure if this is the forum to ask but since its a ALGO VPN thought I'd check anyhow.

jackivanov:
It would be more appropriate to use the name servers which are given by the provider and already configured in the system instead of the ones in the config.
Fixes #806

### DIFF
--- a/roles/dns_adblocking/templates/dnsmasq.conf.j2
+++ b/roles/dns_adblocking/templates/dnsmasq.conf.j2
@@ -88,7 +88,7 @@
 # You can control how dnsmasq talks to a server: this forces
 # queries to 10.1.2.3 to be routed via eth1
 # server=10.1.2.3@eth1
-{% for host in dns_servers.ipv4 %}
+{% for host in ansible_dns['nameservers']|default(dns_servers['ipv4']) %}
 server={{ host }}
 {% endfor %}
 
--- a/roles/vpn/templates/ipsec.conf.j2
+++ b/roles/vpn/templates/ipsec.conf.j2
@@ -31,7 +31,7 @@ conn %default
 {% if local_dns is defined and local_dns == "Y" %}
     rightdns={{ local_service_ip }}
 {% else %}
-    rightdns={% for host in dns_servers.ipv4 %}{{ host }}{% if not loop.last %},{% endif %}{% endfor %}{% if ipv6_support is defined and ipv6_support == "yes" %},{% for host in dns_servers.ipv6 %}{{ host }}{% if not loop.last %},{% endif %}{% endfor %}{% endif %}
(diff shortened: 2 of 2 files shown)


jwebbstevens:
It looks like Algo is configuring dnsmasq to relay requests to Google's DNS servers by default: https://github.com/trailofbits/algo/blob/d8f0393dd8d9a88fe72dec47f0290e9990e9e125/config.cfg#L32

Is that why I'm seeing Googles DNS servers when I do a DNS Leak test with https://whoer.net/ or https://www.dnsleaktest.com/ ? I also turned off all the telemetry data in Firefox which helped https://support.mozilla.org/en-US/questions/1197144.  I'm going through all my plugins one by one and doing the tests as well.

If this commit solves this that's awesome.  Thanks!

gunph1ld:
> Is that why I'm seeing Googles DNS servers when I do a DNS Leak test with https://whoer.net/ or https://www.dnsleaktest.com/ ?

Yes, that's right. This commits solves the problem.

jwebbstevens:
This is a really dumb question of mine (I'm a noob) but how will i know when this commit is merged into the master?  

I'm happy to help test it if anyone wants. It's easy enough for me to grab the changes and spin up another Digital Ocean server and double check the DNS leaks.

Issue #798, opened 2018-02-04
Deployed this to my server and now all my websites are gone!


julien-c:
Deployed `algo` to my server running Ubuntu 16.04 which hosts a bunch of different websites and services.

They are all inaccessible now – what should I do to fix them?

I guess I want to revert my iptables configuration?

dguido:
Yep, please don’t install this on a server you use for anything else. Local installs are your own responsibility. Use the automated cloud deployment methods!

julien-c:
I guess it wouldn't hurt to make that clearer in the README. Thanks!

paulschreiber:
Same thing happened to me. I had to flush iptables, despite my webserver running on a different IP address than my VPN server.

paulschreiber:
The workaround for this is to edit `/etc/iptables/rules.v4` and open those ports.

See https://github.com/paulschreiber/algo/commit/f806ef9ae041ec4a91e1f6a3d72739fdbf5f47fd in branch [fix/iptables](https://github.com/paulschreiber/algo/tree/fix/iptables).

Issue #651, opened 2017-08-20
blocked my other VPN ?


anarcheuz:
### OS / Environment

Linux ip-172-31-16-71 4.4.0-1030-aws #39-Ubuntu SMP Wed Aug 9 09:43:05 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

### Ansible version

ansible 2.2.0.0
  config file = /home/ubuntu/algo-master/ansible.cfg
  configured module search path = Default w/o overrides

### Summary of the problem
Hi, I just setup Algo on my amazon EC2 instance to find out that my shadowsocks server cannot receive connections anymore, server is up but clients cannot connect to server. I suspect this is due to some firewall / ipsec rules dropping incoming packets ? 
I would like to keep shadowsocks just the time to fully test out algo. Is there a way I can make them co-exist ?

dguido:
Algo rewrites your iptables rules

PR #813, opened 2018-02-27
Warning for standalone non-standalone servers

paulschreiber:
People (including me) are installing this on their own servers (see #651, #798) and finding other services stop working.


dguido:
I want to put this somewhere, but I don't think the current location is the right place. I intended the main walkthrough in the README to only cover the cloud deployment scenarios.

PR #823, opened 2018-03-09
Explicitly create the virtualenv with Python2

berryp:
Those who have Python3 as the default global Python may miss that this project only supports Python2 (I did, myself). I've updated the virtualenv command in the README to create the env with Python2 explicitly.


dguido:
Thanks!

PR #827, opened 2018-03-12
update troubleshooting doc

zconnelly:
I ran into an error that I couldn't find in the troubleshooting doc -- so I thought it'd be good to document why it happened as well as the fix! (see the PR itself for details on the error/fix) I also incidentally removed some trailing spaces so there is a tiny bit of noise in the diff.


dguido:
Thanks! Can you bump it to the bottom? The issues at the top are intended to be the absolute most common ones, like the XCode license stuff.

zconnelly:
@dguido Bumped :D

dguido:
Thank you!

PR #842, opened 2018-03-22
Update client-android.md

tightwadjock:
Changed Installation via profiles sections - Opening the helper html file in Chrome (v65.0.3325.109 on Android 6.0.1) does not work correctly.


dguido:
Thanks!

PR #862, opened 2018-04-01
Update config.cfg

LewisLebentz:
Use Cloudflare's new privacy focused DNS resolver by default.

According to DNSPerf, it's the fastest public resolver in the world right now too. (https://www.dnsperf.com/#!dns-resolvers)

### DIFF
--- a/config.cfg
+++ b/config.cfg
@@ -31,11 +31,11 @@ adblock_lists:
 
 dns_servers:
   ipv4:
-    - 8.8.8.8
-    - 8.8.4.4
+    - 1.1.1.1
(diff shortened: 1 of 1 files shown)


dguido:
Thanks, but I'm not interested in changing the default unless someone can describe the measurable benefits of Cloudflare over Google DNS.

LewisLebentz:
@dguido what do you think about having Google's DNS as Primary, with 1.1.1.1 as the Secondary?

If for some reason Google's was blocked, or went down it could use Cloudflare's as a backup.

PR #863, opened 2018-04-02
skip virtualenv check if already activated

zigg:
This allows the user to choose their virtualenv method, e.g. [Pipenv](https://docs.pipenv.org/) while still being friendly for those not familiar with virtual environments.

Pipenv (and other virtual environment management tools) may create the environment somewhere other than inside the project directory.

Note that this solution may be insufficient if the project ever uses Python 3 and supports [venv](https://docs.python.org/3/library/venv.html); we may need to use a Python-based solution then, e.g. https://stackoverflow.com/a/42580137

### DIFF
--- a/algo
+++ b/algo
@@ -2,13 +2,16 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


dguido:
Thanks!

PR #875, opened 2018-04-17
DNS encryption

### LINKED ISSUE #871: DNS over HTTPS
y/n option for routing DNS over HTTPS

[Dingo](https://github.com/pforemski/dingo) can do it with Google DNS, or we can go the route that [Cloudflare did](https://blog.cloudflare.com/dns-over-tls-for-openwrt/) with DNSmasq and Unbound.

jackivanov:
- [dnscrypt-proxy](https://github.com/jedisct1/dnscrypt-proxy) service for supported *BSDs and Ubuntu.
- AppArmor policies and cgroups limits
- CloudFlare DNS over HTTPS enabled by default for all installations
- Fixes #871

### DIFF
--- a/roles/dns_adblocking/tasks/ubuntu.yml
+++ b/roles/dns_adblocking/tasks/ubuntu.yml
@@ -7,13 +7,13 @@
     owner: root
     group: root
     mode: 0600
-  when: apparmor_enabled is defined and apparmor_enabled == true
+  when: apparmor_enabled|default(false)|bool == true
   notify:
     - restart dnsmasq
 
 - name: Ubuntu | Enforce the dnsmasq AppArmor policy
   shell: aa-enforce usr.sbin.dnsmasq
-  when: apparmor_enabled is defined and apparmor_enabled == true
+  when: apparmor_enabled|default(false)|bool == true
   tags: ['apparmor']
 
 - name: Ubuntu | Ensure that the dnsmasq service directory exist
--- a/roles/vpn/tasks/ubuntu.yml
+++ b/roles/vpn/tasks/ubuntu.yml
@@ -12,7 +12,7 @@
 
 - name: Ubuntu | Enforcing ipsec with apparmor
   shell: aa-enforce "{{ item }}"
-  when: apparmor_enabled is defined and apparmor_enabled == true
+  when: apparmor_enabled|default(false)|bool == true
   with_items:
     - /usr/lib/ipsec/charon
     - /usr/lib/ipsec/lookip
--- a/algo
+++ b/algo
@@ -40,7 +40,7 @@ read -p "
 Do you want to install a DNS resolver on this VPN server, to block ads while surfing?
 [y/N]: " -r dns_enabled
 dns_enabled=${dns_enabled:-n}
-if [[ "$dns_enabled" =~ ^(y|Y)$ ]]; then ROLES+=" dns"; fi
+if [[ "$dns_enabled" =~ ^(y|Y)$ ]]; then ROLES+=" dns"; EXTRA_VARS+=" local_dns=true"; fi
 
 read -p "
 Do you want each user to have their own account for SSH tunneling?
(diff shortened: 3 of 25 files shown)


roles/dns_adblocking/tasks/ubuntu.yml
```diff
@@ -7,13 +7,13 @@
     owner: root
     group: root
     mode: 0600
-  when: apparmor_enabled is defined and apparmor_enabled == true
+  when: apparmor_enabled|default(flase)|bool == true
```
jplock: `false`, same comment below

roles/vpn/tasks/ubuntu.yml
```diff
@@ -12,7 +12,7 @@
 
 - name: Ubuntu | Enforcing ipsec with apparmor
   shell: aa-enforce "{{ item }}"
-  when: apparmor_enabled is defined and apparmor_enabled == true
+  when: apparmor_enabled|default(flase)|bool == true
```
jplock: `false`

dguido:
Let's put this in the default configuration, rather than behind the ad blocking role.

I assume that many users may want to change between Google and Cloudflare for DNS-over-HTTPS. Is the setting for that exposed in config.cfg?

dguido:
You know what, let's switch the default to Cloudflare DNS. Cloudflare deletes a larger amount of data from their logs than Google does:

Compare:

* https://developers.google.com/speed/public-dns/privacy
* https://developers.cloudflare.com/1.1.1.1/commitment-to-privacy/privacy-policy/privacy-policy/

dguido:
Yes, skip the y/n prompt. Let’s make this the default for everyone.

dguido:
What does Unbound do? Do we need it or can we just use Dingo?

jackivanov:
Unbound for CloudFlare. Dingo [doesn't support](https://github.com/pforemski/dingo/pull/41) CloudFlare.

beachfork:
If Clouflare only would be an option, cloudflared (argo-tunnel) for DNS-over-HTTPS would be easy to integrate with dnsmasq.

https://developers.cloudflare.com/1.1.1.1/dns-over-https/cloudflared-proxy/

jackivanov:
@beachfork Any benefits compared to Unbound? I don't see a binary for *BSDs

beachfork:
Oh you're right, there's no *BSDs support for now. As far as I know there is DNS-over-TLS and DNS-over-HTTPS (DoH), cloudflared is for DoH. The Unbound method will get you DNS-over-TLS.

jackivanov:
Right. I'd prefer DNS-over-TLS. At least it's RFCed. And should be faster in theory

beachfork:
From benchmarks over the internet DNS-over-TLS seems to be faster! So you're right, Unbound is the universal route.

dguido:
Performance should be similar between HTTPS and TLS for our setup. The AlgoVPN server is doing the lookup, and it will do it from a VPS datacenter over a fast, wired connection. I think it's fine to support _only_ DNS-over-HTTPS.

There is a draft RFC. It requires HTTP2 with strict cipher suites, so it's pretty good. https://www.ietf.org/id/draft-ietf-doh-dns-over-https-07.txt

Let's ditch the Unbound client. I am concerned that it adds too much complexity and it's another random pile of C that we have to stay concerned about. I'm much happier with a simple Go binary with Dingo.

dguido:
Damn it. I just noticed that Dingo may have trouble with Cloudflare DNS. Let's use dnscrypt-proxy instead: https://github.com/jedisct1/dnscrypt-proxy

ndtob:
Maybe it's too late but Stubby https://dnsprivacy.org/wiki/display/DP/About+Stubby is what I used for DNS-over-TLS local resolver. Just throwing it out there. Supports IPv4/6 etc.

dguido:
Niceeee. Let me know when this is ready for review! I'll deploy it and test it.

jackivanov:
Rebased

dguido:
Ehhh never mind. Let’s get it out there quickly.

PR #899, opened 2018-04-27
Update dnscrypt-proxy.toml.j2

adamluk:
Updated dnscrypt-proxy.toml with new options: cache_neg_min_ttl and cache_neg_max_ttl

These new options in Version 2.0.11 are to clamp the negative caching TTL. The repository Algo is using (ppa:shevchuk/dnscrypt-proxy) has updated to this version. These options are now included as default in dnscrypt-proxy.


dguido:
Thanks!

PR #905, opened 2018-04-29
Don't download minisig dnscrypt release

TheNeuralBit:
When installing on FreeBSD, I ran into an issue where a minisig of dnscrypt was being downloaded rather than the actual release. This patch resolved it for me.


dguido:
Thanks!

PR #910, opened 2018-04-30
Add WireGuard support for Android

### LINKED ISSUE #893: Add Wireguard
- Add Wireguard to the default install of Algo
- Make Wireguard the default method for Android users
- Stop making sswan and helper files for IPSEC for Android

jackivanov:
- Adds WireGuard support
- Removes android StrongSwan profiles and helpers
- Fixes #893
- FreeBSD is not supported


vadiof:
This feature is highly expected!

zx2c4:
It's worth noting that now WireGuard runs on macOS too: https://lists.zx2c4.com/pipermail/wireguard/2018-May/002854.html

davidemyers:
I don't have Android but I thought I'd give this a try with Ubuntu Server 18.04 on both ends so I deployed to DigitalOcean.

After installing `resolvconf` on the client everything seems to work fine.

Perhaps WireGuard should be added to `unattended-upgrades`.

dguido:
I need to find an Android device to test this out with. I think the todo list is:

- [ ] Review documentation for accuracy
- [ ] AppArmor profile
- [ ] Add WireGuard to unattended upgrades

davidemyers:
If you want to include documentation for Ubuntu Server here's what I did on 18.04. It was a lot easier than dealing with strongSwan.

Perhaps there's a more proper way to change the DNS settings when `systemd-resolved` is in use, but I find `systemd` somewhat opaque so I just installed `resolvconf`.

I'm not 100% sure there aren't any IPv6 MTU issues lurking for those like me who have to use PPP for our Internet connections.

Your documentation should probably mention that .conf files shouldn't be shared between users like you can do with IPsec configs (when `uniqueids=never`).

```
sudo add-apt-repository ppa:wireguard/wireguard
sudo apt install wireguard resolvconf
# resolvconf did not start after installation for some reason
sudo systemctl start resolvconf
sudo cp wireguard/<username>.conf /etc/wireguard/wg0.conf
sudo chown root:root /etc/wireguard/wg0.conf
sudo chmod 600 /etc/wireguard/wg0.conf
sudo wg-quick up wg0
```

1mentat:
I was able to use this branch with "brew install wireguard-tools", then "./algo", then "wg-quick up ./configs/<ip>/wireguard/<user>.config". Very nice.

zx2c4:
> wg-quick up ./configs/<ip>/wireguard/<user>.config

The standard naming convention for wg-quick configurations is ending in `.conf`. If you put these files into /etc/wireguard -- as /etc/wireguard/burrito.conf, for example -- then you can simply run `wg-quick up burrito`.

1mentat:
It is `<user>.conf` in the PR, I mistyped it. `/etc/wireguard/` won't work on OS X?

zx2c4:
It will work on macOS just fine. Probably best to stay focused on the PR here though. Happy to discuss on #wireguard on Freenode.

jackivanov:
I'm not really sure that we can do anything with apparmor here. Suggestions?

davidemyers:
I've changed my suggested instructions for Ubuntu Server 18.04 since I found I could create DNS leaks using `resolvconf` or `openresolv` with `wg-quick`. Rather than installing either of these packages, one can apparently tell `systemd-resolved` to do the right thing by first editing the generated `<username>.conf` file and replacing:
```
DNS = 172.16.0.1
```
with:
```
PostUp = systemd-resolve -i %i --set-dns=172.16.0.1 --set-domain=~.
```

The setup steps without `resolvconf` become (with some steps combined by using `install`):
```
sudo add-apt-repository ppa:wireguard/wireguard
sudo apt install wireguard
sudo install -o root -g root -m 600 wireguard/<username>.conf /etc/wireguard/wg0.conf
sudo wg-quick up wg0
```

zx2c4:
@davidemyers Happy to discuss the state of resolvconf and systemd-resolved, and systemd 239's resolvconf compat code, with you on the mailing list or private email or IRC (#wireguard) if you want, but not here, because I don't see its relevance to the PR.

@jackivanov Indeed not much to do with apparmor, except perhaps protecting /etc/wireguard, where the private keys live. I saw you added this to unattended-upgrades. While this might succeed at building a new kernel module, after it runs you probably need to do something like `systemctl stop wg-quick@wg0 && rmmod wireguard && systemctl start wg-quick@wg0`, in order to have it load the newly built module. But maybe that's out of scope of unattended upgrades here?

jackivanov:
@zx2c4 protecting /etc/wireguard: I think file permissions would be enough? (no apparmom needed). As for the upgrade, I think such interactions have to be done in the particular deb scripts (eg: postinst) and provided by the maintainer.

README.md
```diff
@@ -97,7 +97,7 @@ Certificates and configuration files that users will need are placed in the `con
 
 ### Android Devices
 
-No version of Android supports IKEv2. Install the [strongSwan VPN Client for Android 4 and newer](https://play.google.com/store/apps/details?id=org.strongswan.android). Import the corresponding user.p12 certificate to your device. See the [Android setup instructions](/docs/client-android.md) for more a more detailed walkthrough.
+WireGuard is used to provide VPN services since IKEv2 is not supported on Android. Install the [WireGuard VPN Client](https://play.google.com/store/apps/details?id=com.wireguard.android). Import the corresponding `<name>.conf` file to your device, then setup a new connection with it. See the [Android setup instructions](/docs/client-android.md) for more detailed walkthrough.
```
omidfi: Shouldn't the file be wireguard directory? 
`wireguard/{username}.conf`

omidfi: By the way it worked perfectly on DigitalOcean, using android 7.0

dguido:
The documentation is pretty ok at this point and it sounds like AppArmor may be unnecessary, though we could add a tiny policy that just covers the files in /etc/wireguard.

However, unattended-upgrades do not work the way we expect it (https://github.com/trailofbits/algo/pull/910#issuecomment-390725900). @zx2c4: Is there any chance you can add a postinst step to the deb to handle the systemctl incantations required to load a new module?

Once we get automatic updates resolved, we should be good to merge this.

zx2c4:
> though we could add a tiny policy that just covers the files in /etc/wireguard.

Indeed, you could restrict reads of those to a certain context, more restricted than just the root user.

>  Is there any chance you can add a postinst step to the deb to handle the systemctl incantations required to load a new module?

Yes. I'm not the deb maintainer, but if you write me a patch for the deb, I'll push it to him. Alternatively, there's a repo at https://github.com/EggieCode/wireguard-ppa you could send a PR for. @EggieCode - can you push your latest to that repo?

The postinst script should probably be something like:

```
#!/bin/sh

[[ -f /etc/wireguard/.reload-module-on-update ]] || exit

units="$(systemctl list-units --all --state=loaded --plain --no-legend --no-pager 'wg-quick@*.service'|cut -d ' ' -f 1)"

for unit in $units; do
	systemctl stop "$unit"
done

rmmod wireguard

for unit in $units; do
	systemctl start "$unit"
done
```

Also CC'ing @dkg, who maintains the Debian package, which is usually what the Ubuntu maintainer copies.

EggieCode:
The git repo has been updated. I quite like this addition. 
I can add that script in here: https://github.com/EggieCode/wireguard-ppa/blob/master/debian/wireguard-dkms.postinst, with a nice post message.

zx2c4:
@EggieCode I'll submit a patch for the updated repo to add it to the existing postinst.

zx2c4:
This is now addressed in https://github.com/EggieCode/wireguard-ppa/pull/28 , though that commit is not tested at all, so please give this a whirl and let me know if it does something remotely functional.

zx2c4:
The new .deb has now been released, so from upstream's end, I think we're done.

README.md
```diff
@@ -97,7 +97,7 @@ Certificates and configuration files that users will need are placed in the `con
 
 ### Android Devices
 
-No version of Android supports IKEv2. Install the [strongSwan VPN Client for Android 4 and newer](https://play.google.com/store/apps/details?id=org.strongswan.android). Import the corresponding user.p12 certificate to your device. See the [Android setup instructions](/docs/client-android.md) for more a more detailed walkthrough.
+WireGuard is used to provide VPN services since IKEv2 is not supported on Android. Install the [WireGuard VPN Client](https://play.google.com/store/apps/details?id=com.wireguard.android). Import the corresponding `wireguard/<name>.conf` file to your device, then setup a new connection with it. See the [Android setup instructions](/docs/client-android.md) for more detailed walkthrough.
```
zx2c4: Maybe get rid of that "since IKEv2 is not supported on Android", since that seems to imply you'd prefer to use IPsec if you had the option, which I sort of hope isn't the case.

docs/client-android.md
```diff
@@ -2,48 +2,6 @@
...
-[step4-screen]: https://i.imgur.com/xYjoNNO.png
-[step5-screen]: https://i.imgur.com/4qhKT1Z.png
-[step6-screen]: https://i.imgur.com/MAaQuxH.png
-[step7-screen]: https://i.imgur.com/aT2MPih.png
-[step8-screen]: https://i.imgur.com/gvaKzkh.png
-[step9-screen]: https://i.imgur.com/eZp8DNb.png
-[step10-screen]: https://i.imgur.com/Nd8rYMJ.png
+1. [Install the WireGuard VPN Client](https://play.google.com/store/apps/details?id=com.wireguard.android).
+2. Copy `wireguard/{username}.conf` to your phone's internal storage.
+3. Open the WireGuard app and create a connection from your configuration file.
```
zx2c4: s/create a connection/add a configuration/

roles/wireguard/tasks/keys.yml
```diff
@@ -0,0 +1,60 @@
+---
+- name: Delete the lock files
+  file:
+    dest: "/etc/wireguard/private_{{ item }}.lock"
+    state: absent
+  when: easyrsa_reinit_existent|bool == True
```
zx2c4: WireGuard doesn't use easyrsa (it doesn't even use the rsa algorithm!), so you may want to rename this variable.

roles/wireguard/templates/server.conf.j2
```diff
@@ -0,0 +1,18 @@
+[Interface]
+Address = {{ wireguard_network_ipv4['subnet'] }}/{{ wireguard_network_ipv4['prefix'] }}{% if ipv6_support %},{{ wireguard_network_ipv6['gateway'] }}/{{ wireguard_network_ipv6['prefix'] }}
+{% endif %}
+
+DNS = {{ local_service_ip }}
+ListenPort = {{ wireguard_port }}
+PrivateKey = {{ lookup('file', wireguard_config_path + '/private/' + IP_subject_alt_name) }}
+SaveConfig = false
```
zx2c4: You may want to change this to true, so that when restarting, if a packet hits the server to be forwarded to a client, the server will re-initiate the session with the client using the client's recent ephemeral port if it's still open.

roles/wireguard/tasks/keys.yml
jackivanov: Probably yes, but it's not related to this PR.

roles/wireguard/templates/server.conf.j2
jackivanov: Any saved config will be rewritten is someone want to update users on the servers. Configuration management is done via Ansible

zx2c4: That's fine. The purpose of adding saveconfig=true is so that when an update brings down the interface and brings it back up, the ephemeral endpoints (and therefore the connections) are restored, and the user's downloads continue as if nothing happened.

davidemyers:
I'm not sure if this is a problem, but I thought I should mention it.

When `unattended-upgrades` ran on my server this morning the updated package was installed and the module was rebuilt but the service was not restarted by the new script, presumably because the module version number did not change.

`systemctl status wg-quick@wg0.service` complained: `Warning: The unit file, source configuration file or drop-ins of wg-quick@wg0.service changed on disk. Run 'systemctl daemon-reload' to reload units.`

My client still had connectivity.

zx2c4:
That sounds to me like intended behavior. The package you received didn't have any changes related to the module, so the module didn't need to be reinserted. Rather, the change it has was one for unattended upgrades of future modules.

.travis.yml
```diff
@@ -44,7 +47,7 @@ env:
   - LXC_NAME=docker LXC_DISTRO=ubuntu LXC_RELEASE=artful
 
 before_install:
-  - test "${LXC_NAME}" != "docker" || docker build -t travis/algo .
+  - test "${LXC_NAME}" != "docker" && sudo modprobe wireguard || docker build -t travis/algo .
```
zx2c4: Why do you modprobe wireguard directly? wg-quick calls `ip link add wg0 type wireguard`, and ip-link calls RTM_NEWLINK and that then calls request_module which finally calls modprobe itself.

davidemyers:
OK, I wasn't sure if that was correct because DKMS completely removed and rebuilt the module even though the version number didn't change.

```
-------- Uninstall Beginning --------
Module:  wireguard
Version: 0.0.20180519
Kernel:  4.15.0-22-generic (x86_64)
-------------------------------------

Status: Before uninstall, this module version was ACTIVE on this kernel.

wireguard.ko:
 - Uninstallation
   - Deleting from: /lib/modules/4.15.0-22-generic/updates/dkms/
 - Original module
   - No original module was found for this module on this kernel.
   - Use the dkms install command to reinstall any previous module version.

depmod...

DKMS: uninstall completed.

------------------------------
Deleting module version: 0.0.20180519
completely from the DKMS tree.
------------------------------
```

.travis.yml
jackivanov: This is for to bridge the module inside LXC containers. We don't start wg-quick on the host system, only inside containers.

zx2c4:
I know. I didn't misunderstand you. If the package needs to make changes to the module on a package bump when there's no upstream version bump, it can easily amend the module's version string before doing so.

.travis.yml
zx2c4: Doesn't `request_module` usually execute processes in the init namespace though? (Did you try this without that line?) Mostly just curious...

jackivanov: I'm not very familiar with LXC, to load the module on the host is the standard recommendation

zx2c4: :+1:

roles/wireguard/tasks/keys.yml
zx2c4: :+1:

zx2c4:
The PR looks good to me at this stage. Ship it!

radiosilence:
How do we update our existing algo VPN servers, or do we have to re-make them every time there's an update?

jackivanov:
@radiosilence rebuild is the best way. You may try to run algo with the same parameters, but at your own risk

PR #912, opened 2018-04-30
Booleans declaration and update users fix

### LINKED ISSUE #885: OnDemand rules is lost when using ./algo update-users
### OS / Environment (where do you run Algo on)
```
Linux vpn1-algo-am 4.4.0-119-generic #143-Ubuntu SMP Mon Apr 2 16:08:24 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

### Cloud Provider (where do you deploy Algo to)
local server

### Summary of the problem
Expected user.mobileconfig to have 
```
<key>OnDemandEnabled</key>
                <integer>1</integer>
              <key>OnDemandRules</key>
              <array>
                  <dict>
                    <key>Action</key>
                      <string>Connect</string>
                      <key>InterfaceTypeMatch</key>
                      <string>WiFi</string>
                    <key>URLStringProbe</key>
                      <string>http://captive.apple.com/hotspot-detect.html</string>
                  </dict>
                  <dict>
                    <key>Action</key>
                      <string>Connect</string>
                      <key>InterfaceTypeMatch</key>
                      <string>Cellular</string>
                    <key>URLStringProbe</key>
                      <string>http://captive.apple.com/hotspot-detect.html</string>
                  </dict>
                </array>
```
after running `./algo update-users`

But user.mobileconfig doesn't include this part

1. Clear install algo from git
2. run `./algo` and answer YES to all OnDemand questions -> now we have OnDemand section in .mobileconfig files
3. add new user to config.cfg
4. run `./algo update-users` -> No questions for OnDemand rule and now OnDemand section on .mobileconfig files

jackivanov:
- Removes unneeded playbook - `playbooks/local_ssh.yml`
- Makes more proper boolean variables declaration
- Adds a dump of all the config options used - `configs/$server_ip/config.yml`
- Simplifies the update-users script (Asking for the CA password and the server IP only now. all the other options are taken from the dump above)
- Fixes #885


dguido:
It looks like this has a few merge conflicts!

jackivanov:
I will cover it in another PR

Issue #887, opened 2018-04-23
SecretStorage requires Python '>=3.5' but the running Python is 2.7.12


chekalskiy:
### OS / Environment (where do you run Algo on)
```
Linux 4.4.0-116-generic #140-Ubuntu SMP Mon Feb 12 21:23:04 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

### Summary of the problem
When I am installing Algo to my own Ubuntu server i've got an error.

[New SecretStorage](https://pypi.org/project/SecretStorage/#history) version breaks the installation.

### Steps to reproduce the behavior
1. Install Ubuntu 16.04
2. Do as described [here](https://github.com/trailofbits/algo/blob/master/docs/deploy-to-ubuntu.md)
3. Do the line `python -m virtualenv env && source env/bin/activate && python -m pip install -U pip && python -m pip install -r requirements.txt`
4. Get the error

### Full log
```
Collecting secretstorage; sys_platform == "linux2" or sys_platform == "linux" (from keyring>=5.6->msrestazure<0.5,>=0.4.11->ansible[azure]==2.4.3->-r requirements.txt (line 2))
  Downloading https://files.pythonhosted.org/packages/02/de/e4590cf025c714635637d4b2d19acfb1015f2d0c6cfe843b2b8f776ec819/SecretStorage-3.0.0-py2.py3-none-any.whl
```

movalex:
I just added `SecretStorage < 3` before `ansible[azure]==2.4.3` in `requirements.txt`

chekalskiy:
@movalex yep, that works. Maybe you could create PR to fix this issue?

cgm616:
This should be closed now.

PR #914, opened 2018-04-30
fix requirements.txt SecretStorage version

movalex:
Related to issue #887. Latest SecretStorage build requires Python '>=3.5' but Algo is running on Python 2


dguido:
@jackivanov Is this good to merge?

jackivanov:
Yes

PR #928, opened 2018-05-08
Add new EC2 regions

pguizeline:
Adds new EC2 regions according to:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions


dguido:
Thanks!

PR #929, opened 2018-05-08
Add new Azure locations

pguizeline:
Reorganized and added new locations.
https://azure.microsoft.com/en-us/global-infrastructure/locations/
https://azure.microsoft.com/en-us/global-infrastructure/services/


dguido:
Thanks!

PR #931, opened 2018-05-08
Update README.md

pguizeline:
- Adds missing providers to the documentation with links.
- Mentions that your own server install needs to be an Ubuntu 16.04 LTS distro
- Emphasize that the p12 certificate password will only be available once


dguido:
Thank you!

PR #936, opened 2018-05-09
Update config.cfg

TC1977:
Fix typos - this puzzled me when I was attempting to install algo with dnscrypt last week.


dguido:
Thanks!

PR #943, opened 2018-05-13
Arch Linux server support

gagarski:
Tested on new DO Droplet with Debian 9.4 converted to Arch Linux with [this script](https://github.com/gh2o/digitalocean-debian-to-arch)


roles/dns_encryption/tasks/arch.yml
```diff
@@ -0,0 +1,13 @@
...
+
+- name: Arch Linux | Install dnscrypt-proxy
+  pacman:
+    name: dnscrypt-proxy
+    state: latest
+
+- name: Arch Linux | Patch dnscrypt-proxy.service
+  template:
+    src: dnscrypt-proxy.service.j2
+    dest: "{{ config_prefix|default('/') }}etc/systemd/system/dnscrypt-proxy.service"
```
gagarski: I don't like this part very much.

Here I patch the service file to disable SystemD socket activation which Arch package provides by default.

I did it because I coulnd't make `dnscrypt-proxy.socket` start **after** activating `172.16.0.1` loopback interface. Though if I maneged to do it I still had to patch `dnscrypt-proxy.socket` to change the listen address.

roles/dns_adblocking/tasks/arch_set_nogroup.yml
```diff
@@ -0,0 +1,5 @@
+---
```
gagarski: Perhaps I could've moved this part to `arch.yml` but then I'd have to rearrange items in `main.yml`.

jackivanov:
I'm not really sure that we are going to support more systems

roles/dns_encryption/tasks/arch.yml
gagarski: I am not sure if a dependency on `dnscrypt-proxy.socket` can be removed this way.

gagarski:
@jackivanov 

I see your point. It's quite a burden to maintain multiple systems and especially Arch Linux. So feel free to close it. Maybe sometime some Arch-maniac like me will find this PR among closed ones and will use it to turn his Archlinux server into algo server.

PS For some reason deploying Archlinux client turned out to be more hard for me.

PR #957, opened 2018-05-22
 Update GCP regions

sbalk:
Update GCP regions according to https://cloud.google.com/compute/docs/regions-zones/. Change default choice to region 20: Belgium B.


dguido:
Thanks! Please leave the default region as-is and I'm happy to merge it.

sbalk:
Meant to say 'Change default choice back to region 20: Belgium B. So the default region is still as-is but moved to a different position 14 -> 20.

dguido:
haha I did not even realize we set Belgium as the default. Oops. I guess this is fine then! Thanks!

PR #967, opened 2018-05-25
fix faq entry about cryptography build failure

reaperhulk:
The error in the FAQ is actually a `cryptography` build failure. Any macOS user on `pip` > 8.1 will not see this though since the project builds binary wheels now.


dguido:
Thanks!

PR #968, opened 2018-05-25
Extra line and better DNS configuration for WireGuard

### LINKED ISSUE #965: Wireguard giving error
### OS / Environment (where do you run Algo on)

```
Linux  4.15.0-1008-gcp #8-Ubuntu SMP Thu May 17 08:10:14 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

### Cloud Provider (where do you deploy Algo to)

```
Google
```

### Summary of the problem
I'm trying to setup my wireguard client. 
Used latest build of algo (git clone)
latest version of android on a pixel 8.1.0

1. Download /wireguard/user.conf onto android
2. Open in wireguard
3. Wireguard gives error

### Full log
"Unable to import tunnel: not a numberic address: xx.xx.xx.xxx/32/DNS = xxx.xx.x.x
screenshot may 24 2018 4_52_42 pm

jackivanov:
- Adds an extra line after the if statement. Jinja2 trims such blocks by default in Ansible. Fixes #965
- More appropriate way to configure DNS servers
- Removes `DNS` option from the wireguard server config
- Fixes dnscrypt-proxy restart

### DIFF
--- a/roles/common/tasks/ubuntu.yml
+++ b/roles/common/tasks/ubuntu.yml
@@ -89,7 +89,6 @@
       - iptables-persistent
       - cgroup-tools
       - openssl
-      - resolvconf
     sysctl:
       - item: net.ipv4.ip_forward
         value: 1
--- a/roles/dns_encryption/handlers/main.yml
+++ b/roles/dns_encryption/handlers/main.yml
@@ -4,6 +4,7 @@
     daemon_reload: true
 
 - name: restart dnscrypt-proxy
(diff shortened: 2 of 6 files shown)


davidemyers:
Is the DNS setting necessary in the `server.conf`? It looks like it doesn't get saved by `SaveConfig`.

jackivanov:
Yep, we don't really need it

PR #971, opened 2018-05-27
Auto create client network manager config using ansible

seamustuohy:
This pull request removes the need for Linux users to use the Network Manager GUI to create the VPN config. The shell script that is added uses the Network Manager CLI to create a VPN connection profile at the end of the deploy_client.yml Ansible playbook.

### DIFF
--- a/library/linux_nm_client_install.sh
+++ b/library/linux_nm_client_install.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+
+# Terminate if command or chain of command finishes with a non-zero exit status.
+set -e
+# Terminate if an uninitialized variable is accessed.
+set -u
+
+main() {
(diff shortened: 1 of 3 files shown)


jackivanov:
You can avoid using bash scripts with [this](http://docs.ansible.com/ansible/latest/modules/nmcli_module.html) don't hardcode ipsec options, please

seamustuohy:
I didn't see the functionality to add arbitrary VPN options in the nmcli ansible module. It seemed to high level.

seamustuohy:
It also seems that the ciphers are only defined in the VPN role. Extracting them would require the client role to read from the VPN role. Quite an anti-pattern. I'll close this pull request down since this would require some heavy restructuring that I don't have time for.

PR #925, opened 2018-05-07
Ubuntu1804

jackivanov:
- Fixes #897 #944 #956

Work in progress. Lightsail is not ready for Ubuntu 18.04 yet

- [x] DigitalOcean
~~- [ ] Amazon Lightsail~~
- [x] Amazon EC2
- [x] Microsoft Azure
- [x] Google Compute Engine
- [x] Scaleway
- [x] OpenStack (DreamCompute optimised)

PR #982, opened 2018-05-31
[WIP] Update README.md

dguido:


michael-myers:
One of the things this PR deletes is a mention of Amazon Lightsail.

Was there a plan to remove support for Lightsail? I saw something recently about people using it again because of a new lower price point.

dguido:
Lightsail doesn’t support 18.04 so there’s nothing I can really do. When they do support 18.04, I’ll add them back to the readme.

michael-myers:
I see now. This PR can probably be deleted then since it looks like these changes were made in https://github.com/trailofbits/algo/pull/925

PR #992, opened 2018-06-04
Update troubleshooting.md

TC1977:
Many times people are reaching VPC limits not because they're running other VPCs on AWS, but because they've already deployed several times (AWS allows five VPCs per region). This lets people know they can simply delete their old VPCs instead of contacting AWS support.


dguido:
Thanks!

Issue #1010, opened 2018-06-24
Internet appears offline once connected to the deployed VPN


wiiind:
### OS / Environment (where do you run Algo on)

```
Darwin 17.6.0 Darwin Kernel Version 17.6.0: Tue May  8 15:22:16 PDT 2018; root:xnu-4570.61.1~1/RELEASE_X86_64 x86_64
```

### Cloud Provider (where do you deploy Algo to)

```
Azure
```

### Summary of the problem

Process of deployment seems successful. Screen shows success message with password and IP. But once connected to the deployed VPN, internet appears offline. Internet works if I ssh into the server and ping an address, but the same address appears offline if I just ping it in a local terminal.

### Steps to reproduce the behavior

1. Deploy
2. Connect VPN
3. Internet goes offline(Mac+Windows 10+iOS)

### Full log

```
What provider would you like to use?
    1. DigitalOcean
    2. Amazon EC2
    3. Microsoft Azure
    4. Google Compute Engine
    5. Scaleway
    6. OpenStack (DreamCompute optimised)
    7. Install to existing Ubuntu 16.04 server (Advanced)

Enter the number of your desired provider
: 3

Enter your azure secret id (https://github.com/trailofbits/algo/blob/master/docs/cloud-azure.md)
You can skip this step if you want to use your defaults credentials from ~/.azure/credentials
[pasted values will not be displayed]
[...]: 

Enter your azure tenant id (https://github.com/trailofbits/algo/blob/master/docs/cloud-azure.md)
You can skip this step if you want to use your defaults credentials from ~/.azure/credentials
[pasted values will not be displayed]
[...]: 

Enter your azure client id (application id) (https://github.com/trailofbits/algo/blob/master/docs/cloud-azure.md)
You can skip this step if you want to use your defaults credentials from ~/.azure/credentials
[pasted values will not be displayed]
[...]: 

Enter your azure subscription id (https://github.com/trailofbits/algo/blob/master/docs/cloud-azure.md)
You can skip this step if you want to use your defaults credentials from ~/.azure/credentials
[pasted values will not be displayed]
[...]: 

Name the vpn server:
[algo]: algoAzuEaUsII


  What region should the server be located in? (https://azure.microsoft.com/en-us/regions/)
    1.  East US              (Virginia)
    2.  East US 2            (Virginia)
    3.  Central US           (Iowa)
    4.  North Central US     (Illinois)
    5.  South Central US     (Texas)
    6.  West Central US      (Wyoming)
[...]

PR #1011, opened 2018-06-25
Systemd Relosved

jackivanov:
Move DNSCrypt proxy `fallback_resolver` to systemd-resolved.
Sometimes 1.1.1.1 is unreachable via 53/udp and prevents dnscrypt proxy from start


jackivanov:
Fixes #1010

PR #1019, opened 2018-06-30
Update deploy-to-ubuntu.md

TC1977:
rewrite of #813. Probably better off in the local install how-to than the main readme.


dguido:
Thanks!

PR #1022, opened 2018-07-10
Update dnscrypt-proxy.toml.j2

adamluk:
dnscrypt-proxy version 2.0.16 (available from jedisct1/dnscrypt-proxy) now includes a new config option:

user_name = 'nobody'

On Unix-like systems, the server can run as an unprivileged user,
and the main process will automatically restart if an error occurs.

This pull request enables this function.

### DIFF
--- a/roles/dns_encryption/templates/dnscrypt-proxy.toml.j2
+++ b/roles/dns_encryption/templates/dnscrypt-proxy.toml.j2
@@ -41,6 +41,18 @@ listen_addresses  = ['{{ local_service_ip }}:{{ listen_port }}']
 max_clients = 250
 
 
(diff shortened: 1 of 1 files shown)


dguido:
Neat, thanks!

jackivanov:
Thanks! Let's merge it as an update, but the systemd service unit is supplied with `DynamicUser=yes`

TC1977:
I think this is breaking the server. A lot of people have been complaining on gitter that they can't access websites despite running Algo and connecting to the VPN server successfully. I replicated this with a fresh install onto AWS EC2. When I edit `/etc/dnscrypt-proxy/dnscrypt-proxy.toml` to comment this line out, and then `sudo dnscrypt-proxy service restart`, it works again.

PR #1029, opened 2018-07-19
Run dnsmasq as the dnsmasq user

faf0:
This change lets dnsmasq run as user "dnsmasq" instead of "nobody".
Remove the task that checks whether the dnsmasq user exists.

### DIFF
--- a/roles/dns_adblocking/tasks/main.yml
+++ b/roles/dns_adblocking/tasks/main.yml
@@ -8,9 +8,6 @@
     - name: Dnsmasq installed
       package: name=dnsmasq
 
-    - name: Ensure that the dnsmasq user exists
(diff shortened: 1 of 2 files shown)


roles/dns_adblocking/tasks/main.yml
```diff
@@ -8,7 +8,7 @@
     - name: Dnsmasq installed
       package: name=dnsmasq
 
-    - name: Ensure that the dnsmasq user exist
+    - name: Ensure that the dnsmasq user exists
```
jackivanov: This task should be deprecated and deleted

faf0: Deleted it

dguido:
Thanks!

PR #1038, opened 2018-07-29
Add GCP Los Angeles Region (us-west2)

DeltaSpark:
- Added 1 new region (us-west2) with 3 new zones (a, b, c)
- Changed the default region back to europe-west1-b (see https://github.com/trailofbits/algo/pull/957)
- See https://cloud.google.com/compute/docs/regions-zones/ for new zone documentation

### DIFF
--- a/algo
+++ b/algo
@@ -435,49 +435,52 @@ Name the vpn server:
     8. Western US              (Oregon A)
     9. Western US              (Oregon B)
     10. Western US             (Oregon C)
-    11. Eastern US             (Northern Virginia A)
-    12. Eastern US             (Northern Virginia B)
-    13. Eastern US             (Northern Virginia C)
-    14. Eastern US             (South Carolina B)
-    15. Eastern US             (South Carolina C)
-    16. Eastern US             (South Carolina D)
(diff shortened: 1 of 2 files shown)


dguido:
Thanks! We're working on the Ansible2.5 branch right now, which should auto-populate all the available regions. I'm going to hold this for a little bit. Can you check the ansible2.5 branch and see if these regions are on it?

DeltaSpark:
Ran the Ansible2.5 branch, tried to validate all regions were returning, but got the following error.

`fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'server_name' is undefined\n\nThe error appears to have been in '/Users/*****/algo/roles/cloud-gce/tasks/prompts.yml': line 18, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- set_fact:\n  ^ here\n"}`

That said, looking through the yml files, I'd expect that the regions would auto-populate as expected.

Happy to abandon this pull request if manually adding the regions is not useful for now. Thanks @dguido for the feedback!

PR #1039, opened 2018-07-29
Update PPA for dnscrypt-proxy to 'bionic'

bghost:
This should keep the PPA for dnscrypt-proxy in sync with Ubuntu 18.04.


jackivanov:
Thanks!

PR #1042, opened 2018-07-30
Update troubleshooting docs to include iOS reconnection loop

QuentinMoss:
@jackivanov reference https://github.com/trailofbits/algo/issues/963


dguido:
Thanks!

Issue #1079, opened 2018-08-28
Documented AWS walkthrough user needs additional policies


TC1977:
### OS / Environment (where do you run Algo on)
```
Darwin MyComp.local 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64
```

### Cloud Provider (where do you deploy Algo to)

```
AWS
```

### Summary of the problem
The AWS walkthrough advises creating a user with limited policies (rather than just granting full "AdministratorAccess"). It seems that with the move to ansible2.5, the cloud install onto AWS option fails at the "get regions" step. Not sure why this wasn't failing for me before during the testing phase.

### Steps to reproduce the behavior
1. Create user with limited privileges as per https://github.com/trailofbits/algo/blob/master/docs/deploy-from-ansible.md#minimum-required-iam-permissions-for-deployment
2. Try to install onto AWS.
3. Watch it fail.
4. Create user with full AdministratorAccess.
5. Try to install onto AWS.
6. Watch it succeed.

### Full log
(env) MyComp:algo-master admin$ ./algo
```
PLAY [Ask user for the input] **************************************************

TASK [Gathering Facts] *********************************************************
ok: [localhost]
[pause]
What provider would you like to use?
    1. DigitalOcean
    2. Amazon EC2
    3. Vultr
    4. Microsoft Azure
    5. Google Compute Engine
    6. Scaleway
    7. OpenStack (DreamCompute optimised)
    8. Install to existing Ubuntu 18.04 server (Advanced)
  
Enter the number of your desired provider
:
2

TASK [pause] *******************************************************************
ok: [localhost]

TASK [Set facts based on the input] ********************************************
ok: [localhost]
[pause]
Name the vpn server
[algo]
:
algore

TASK [pause] *******************************************************************
ok: [localhost]
[pause]
Do you want macOS/iOS clients to enable "VPN On Demand" when connected to cellular networks?
[y/N]
:
y

TASK [pause] *******************************************************************
ok: [localhost]
[pause]
Do you want macOS/iOS clients to enable "VPN On Demand" when connected to Wi-Fi?
[y/N]
:
y

TASK [pause] *******************************************************************
ok: [localhost]
[pause]
List the names of trusted Wi-Fi networks (if any) that macOS/iOS clients exclude from using the VPN
[...]

PR #1080, opened 2018-08-28
Update minimum required AWS permissions

TC1977:
Ansible2.5 allows Algo to directly ask AWS for the region list, rather than have it hardcoded and updated manually. Updated the documented minimum required permissions to include "DescribeRegions". Fix for #1079, but people will have to update their existing user permissions manually.


TC1977:
The relevant image on algo/docs/cloud-amazon-ec2.md will need to be updated. @michael-myers?

PR #959, opened 2018-05-23
Fixes for #491 - Naming in mobileconfig, powershell, and AWS EC2

bkono:
Normalized the various cloud_server_name entries to just
algo_server_name. Integrated algo_server_name into the mobileconfig
template as well as powershell. If a default name is provided, the 
original behavior of IP_subject_alt_name will be used to avoid conflicts. 
Also updated the Name tag in the cloudformation template to utilize the 
StackName (which is based on the algo_server_name).

PR #1101, opened 2018-09-03
Fixes #491 by piping algo_server_name through additional templates

### LINKED ISSUE #491: Set the name shown to the user (client) to be the server name specified in the install script
### OS / Environment

AWS

### Summary of the problem
During install a server name can be specified in the script. (A query could also be made for a "friendly name".
Regardless of what is entered, the name used is "Algo"  for:
AWS Console
Apple MobileConfig 
(and probably others that I can't test)

normal install

cloud

### Expected behavior
use the specified server name or optionally a friendly name/tag

bkono:
EC2 Name tag in the cfn stack, VPNName in client_windows.ps1 and the
mobileconfig files all make use of the algo_server_name with a default
of the original Algo VPN naming when the server name is left at default
algo or algo.local.

### DIFF
--- a/roles/cloud-ec2/files/stack.yml
+++ b/roles/cloud-ec2/files/stack.yml
@@ -21,7 +21,7 @@ Resources:
       InstanceTenancy: default
       Tags:
         - Key: Name
-          Value: Algo
+          Value: !Ref AWS::StackName
         - Key: Environment
           Value: Algo
 
@@ -38,7 +38,7 @@ Resources:
         - Key: Environment
           Value: Algo
         - Key: Name
-          Value: Algo
+          Value: !Ref AWS::StackName
 
   Subnet:
     Type: AWS::EC2::Subnet
@@ -49,7 +49,7 @@ Resources:
         - Key: Environment
           Value: Algo
         - Key: Name
-          Value: Algo
+          Value: !Ref AWS::StackName
       VpcId: !Ref VPC
 
   VPCGatewayAttachment:
@@ -66,7 +66,7 @@ Resources:
         - Key: Environment
           Value: Algo
         - Key: Name
-          Value: Algo
+          Value: !Ref AWS::StackName
 
   Route:
     Type: AWS::EC2::Route
@@ -140,7 +140,7 @@ Resources:
           CidrIp: 0.0.0.0/0
       Tags:
         - Key: Name
-          Value: Algo
+          Value: !Ref AWS::StackName
         - Key: Environment
           Value: Algo
 
@@ -181,7 +181,7 @@ Resources:
             cfn-signal -e $? --stack ${AWS::StackName} --resource EC2Instance --region ${AWS::Region}
       Tags:
         - Key: Name
-          Value: Algo
+          Value: !Ref AWS::StackName
         - Key: Environment
           Value: Algo
 
--- a/roles/cloud-ec2/tasks/cloudformation.yml
+++ b/roles/cloud-ec2/tasks/cloudformation.yml
@@ -1,4 +1,3 @@
----
 - name: Deploy the template
   cloudformation:
     aws_access_key: "{{ access_key }}"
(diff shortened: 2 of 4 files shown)


bkono:
This PR is a replacement for PR #959 based off of current master with the ansible 2.5 changes. I've tested against EC2 and confirmed the changes are working there as expected. Open to any feedback or suggested changes.
/cc @TC1977 

mobileconfig
EC2 naming

davidemyers:
This is totally subjective, but how about using "Algo VPN" before the name in both cases and dropping the "IKEv2"?
```
{% if (algo_server_name == 'algo') or (algo_server_name == 'algo.local') %}
  {% set vpn_server_name = 'Algo VPN '+IP_subject_alt_name %}
{% else %}
  {% set vpn_server_name = 'Algo VPN '+algo_server_name %}
{% endif %}
```

bkono:
@davidemyers Thanks for the initial feedback. I'd happily drop the Ikev2 part, that was just carried forward from the existing name. 

For the Algo VPN prefix, at least in my use case, that would unfortunately undo the benefits of this particular naming change for two reasons. 1 - "Algo VPN" + anything, IP included, is truncated with default sizing of the network preferences for macOS on most machines. 2 - For my team, the goal is also to eliminate the repetitive, non-useful portion where we can. As context, I'll generally have ~20 VPNs available at any given time, 5+ being algo provisioned. The fact they are algo servers is actually just noise, since we'd rather scan the list for `company-name-region` or similar. Adding the noise back would mean needing to change the name of every VPN after adding the profile to get to something useful.

davidemyers:
The truncation of names in Network Preferences is unfortunate, but at least there are tooltips that pop up when you mouse over them.

Your case seems atypical. I imagine most users would want to be able to quickly spot their Algo VPN entries in the list.

bkono:
@davidemyers Agreed on my case being atypical, as I've found low uptake of Algo within the devops community so far. Which is a shame btw, since it has a great provisioning experience and is better hardened than most VPNs I've run across. The algo-tf experiment might be a boon to getting better adoption in the community via the module registry.

As an interesting counterpoint, the above two reasons weren't the actual prompt for me finding the original April 2017 issue #491 ... it was actually the conversation that followed after giving out the mobileconfig to a half dozen non-technical folk. Unanimously the conversation followed the pattern of "Whats an Algo...", and then "Couldn't you just name it ...?". I started modifying the mobileconfig, and then ran across the old issue requesting the change when I noticed the server name was inaccurate within the EC2 instance list.

Definitely understand if the desire is to brand / identify the tech in some way. Maybe just a reordering would do the trick? Postfix it as `AlgoServerName Algo VPN` instead? I'd also be interested in hearing from  others, such as @bluetardis who opened the original issue requesting the same change.

bluetardis:
I don't know why more people aren't using it as I have found it a great
solution and probably the easiest I have worked with if you are using an
all "Apple" ecosystem.

Branding isn't a problem - would be happy to have the Algo name in there
(that said, people using it wouldn't generally know what that meant).  For
me it was a usability issue as we have at least 2 and sometimes 3 VPNs to
deal with (eg Sydney/London/East or West Coast US).  Figuring out which one
was what in the naming was a PITA.  I ended up with a script doing a hack
on the mobileconfig which fixed the names but it should have been needed.

PB

bkono:
@bluetardis agreed on ease of use. With the addition of wireguard, I've found it is now the easiest option available for mac/ios, android, chromebooks, and linux. Appreciate the insight on why you opened the original issue... sounds very similar to my situation (trying to spot name-region readily).

@davidemyers Lmk which way you'd like to go with naming. I'm in agreement with @bluetardis that the Algo name is fine to include, but the majority of people receiving the profiles won't know what it means. At least in the dozen or so deployments I've done or been a consumer of, its a 1 to N scenario of deployer to receivers of the profiles. Seems like the algo-server-name Algo VPN naming might be the most straightforward option to get everything in there. Could even titlecase it by default if it looked cleaner/clearer. I'm good with whatever, can always just keep a fork going as I've been doing.

davidemyers:
My opinion hasn't changed, but it's just my opinion. Dan and Jack should make the final decision.

I'm looking forward to the EC2 naming changes. As someone who doesn't have a PhD in EC2 it will help with figuring out what can be safely deleted.

jackivanov:
Leaving it as is for now. It might considered later

PR #1111, opened 2018-09-08
clarify successful install message and iOS install instructions

zoof:
The term, "new users", in the successful server install message is a little confusing as it seems to suggest that the password will be for newly created users.  At least that was my inference when I read it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ Certificates and configuration files that users will need are placed in the `con
 
 ### Apple Devices
 
-**Send users their Apple Profile.** Find the corresponding mobileconfig (Apple Profile) for each user and send it to them over AirDrop or other secure means. Apple Configuration Profiles are all-in-one configuration files for iOS and macOS devices. On macOS, double-clicking a profile to install it will fully configure the VPN (user will be prompted for the p12 password printed following successful server installation). On iOS, users are prompted to install the profile as soon as the AirDrop is accepted.
+**Send users their Apple Profile.** Find the corresponding mobileconfig (Apple Profile) for each user and send it to them over AirDrop or other secure means. Apple Configuration Profiles are all-in-one configuration files for iOS and macOS devices. On macOS, double-clicking a profile to install it will fully configure the VPN. On iOS, users are prompted to install the profile as soon as the AirDrop is accepted. User will be prompted for the appropriate p12 password.
 
 **Turn on the VPN.** On iOS, connect to the VPN by opening Settings and clicking the toggle next to "VPN" near the top of the list. On macOS, connect to the VPN by opening System Preferences -> Network, finding Algo VPN in the left column and clicking "Connect." On macOS, check "Show VPN status in menu bar" to easily connect and disconnect from the menu bar.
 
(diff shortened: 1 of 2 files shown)


davidemyers:
> The term, "new users", in the successful server install message is a little confusing as it seems to suggest that the password will be for newly created users.

I believe that's actually correct, though. For example, if you later run `./algo update-users` the same message will be used to display a different p12 password that applies only to any users newly created by running `./algo update-users`. So changing the message to "all regular users" isn't quite right.

Maybe "user accounts that were just created"? There's probably a better way to put it.

zoof:
What about actually listing the users to which the password pertains?  If you can tell me the appropriate means for referencing the list of users (e.g., `{{ users }}`), I would be happy to modify my PR.

TC1977:
The p12 password is used by MacOS, iOS, and Win10 clients when installing their mobileconfig or running the PowerShell script. I don't know if Ubuntu clients on Strongswan need it (don't have my own Ubuntu machine). I also don't know if Wireguard clients (primarily on Android and Ubuntu, currently) need it.

Maybe I'm just biased, because I just fixed the spacing in the congrats message, but what about mentioning the p12 password more prominently in the Apple and Win10 install instructions instead? I'd move your text to after the iOS sentence, as it applies to both. "New users" on the congrats message seemed pretty straightforward and accurate to me.

zoof:
> The p12 password is used by MacOS, iOS, and Win10 clients when installing their mobileconfig or running the PowerShell script. I don't know if Ubuntu clients on Strongswan need it (don't have my own Ubuntu machine). I also don't know if Wireguard clients (primarily on Android and Ubuntu, currently) need it.

Linux using Strongswan does not need the password, although the install instructions for Debian are slightly different than Ubuntu.  (I have added a line to README.md.)  I have not tried Wireguard on Linux but for Android, it did not ask for a password.  

>Maybe I'm just biased, because I just fixed the spacing in the congrats message, but what about mentioning the p12 password more prominently in the Apple and Win10 install instructions instead? I'd move your text to after the iOS sentence, as it applies to both. "New users" on the congrats message seemed pretty straightforward and accurate to me.

I changed "user will be prompted for the p12 password printed following successful server installation" to "user will be prompted for the appropriate p12 password" and moved it as suggested.  I also copied this parenthetical remark to the Windows 10 section as well.

I have to say that I still disagree but I may be in the minority.  Yes, "new users" could be referring to the just created users but it could also be referring to subsequently created users, which as little sense as it makes now that I know more, is how I read it.  What about "The p12 and SSH keys password for the users just created..."?

jackivanov:
Well, currently,  [this task](https://github.com/trailofbits/algo/blob/master/roles/vpn/tasks/openssl.yml#L136-L150) generates p12  containers with new password each time you update users. The p12 password is new, but the same for all users, each run. So it seems neither "new users" nor "the users just created" appropriate here :)

> Linux using Strongswan does not need the password

It doesn't depend on the software. The password is required for p12 containers only, so if a user carries over the p12 container it does need the password.

jackivanov:
Since last releases we add the p12 password directly to the config files, thus it won't be prompted to enter anything for users. I guess we can eliminate this PR now. Thanks!

PR #1117, opened 2018-09-17
docs/client-android: add instructions to add via QR code

broady:
## Types of changes

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.
- [ ] My code follows the code style of this project.

### DIFF
--- a/docs/client-android.md
+++ b/docs/client-android.md
@@ -3,5 +3,10 @@
 ## Installation via profiles
 
 1. [Install the WireGuard VPN Client](https://play.google.com/store/apps/details?id=com.wireguard.android).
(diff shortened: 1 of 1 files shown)


jackivanov:
Would it be better to create a png and place in the config directory instead? Just to prevent users from executing any commands at all and just open the QR picture and scan it

broady:
Oh, yes, it'd be nice if the QR code was generated by default, but does add another dependency of a QR code library.

For the docs, could add a different command that outputs a PNG and opens it (for desktop systems)
For headless (no desktop), `qrencode` with the terminal output works very well.

jackivanov:
Closing in favor of https://github.com/trailofbits/algo/pull/1129

PR #1120, opened 2018-09-18
Allow more flexible DNSCrypt configuration

davidemyers:
## Description

This PR changes how the encrypted DNS servers are configured in `config.cfg` in order to allow the user to specify any DNSCrypt servers included in the public server list. The servers are now configured the same way that non-encrypted servers are and there is no special code for handling `cloudflare-ipv6`.

I also changed the location of the `cache_file` for the `public-resolvers` list to eliminate the errors that were occurring when `dnscrypt-proxy` starts.

## Motivation and Context

The motivation is to allow a user to easily choose [Quad9](https://quad9.net) as a provider once their DNSCrypt support is out of [beta](https://quad9.net/privacy-dnscrypt-testing). In my testing Quad9 works fine, but to use it during beta an addition to `[sources]` in `dnscrypt-proxy.toml` is required as their servers are not yet in the public servers list. This addition is *not* part of this PR.

If anyone reading this PR wants to try Quad9 while still in beta, add the following to `[sources]` in `roles/dns_encryption/templates/dnscrypt-proxy.toml.j2`:
```
  [sources.quad9-resolvers]
  urls = ["https://www.quad9.net/quad9-resolvers.md"]
  minisign_key = "RWQBphd2+f6eiAqBsvDZEBXBGHQBJfeG6G+wJPPKxCZMoEQYpmoysKUN"
  cache_file = "/run/dnscrypt-proxy/quad9-resolvers.md"
  refresh_delay = 72
  prefix = ""
```

Then use the following in `config.cfg`:
```
dnscrypt_servers:
  ipv4:
#   - cloudflare
#   - google
    - quad9 dnscrypt ip4-filter-pri
  ipv6:
#   - cloudflare-ipv6
    - quad9 dnscrypt ip6-filter-pri
```

## How Has This Been Tested?

I tested the templating code by running Ansible to invoke the `template` module directly. I also deployed several times to DigitalOcean and reviewed the `dnscrypt-proxy` startup logs.

## Types of changes

## Checklist:


davidemyers:
I was expecting Travis to start upon submission, after which I would edit the PR to check off "All new and existing tests passed", but it doesn't look like it has started.

dguido:
The status is not being reported for some reason, but you passed: https://travis-ci.org/trailofbits/algo/builds/430023483

roles/dns_encryption/templates/dnscrypt-proxy.toml.j2
```diff
@@ -446,7 +450,7 @@ cache_neg_max_ttl = 600
 
   [sources.'public-resolvers']
   urls = ['https://raw.githubusercontent.com/DNSCrypt/dnscrypt-resolvers/master/v2/public-resolvers.md', 'https://download.dnscrypt.info/resolvers-list/v2/public-resolvers.md']
-  cache_file = 'public-resolvers.md'
+  cache_file = '/run/dnscrypt-proxy/public-resolvers.md'
```
jackivanov: Wouldn't it break *BSD?

davidemyers: Yes I guess it would, I didn't think of that. Should I add code to prepend `/run/dnscrypt-proxy/` to the file name only in the case of Ubuntu?

jackivanov: Can we use a relative path as before?

davidemyers: I didn't see a way to tell `dnscrypt-proxy` to write these files somewhere other than `/etc/dnscrypt-proxy`, which it can't because it doesn't run as root.

jackivanov: then, I think to add a task, and ensure the directory `/run/dnscrypt-proxy/` exists, would be enough

jackivanov: or just to not make it complex and place the file in `/tmp`

davidemyers: We could use `/tmp`, I just thought using one of the `systemd` created directories was more elegant. :smile: I probably should have chosen `/var/cache/dnscrypt-proxy` anyway.

It seems like the root of this problem is a bug in the PPA package, perhaps.

I'll change it to `/tmp`. I'll do a test deploy before I commit.

jackivanov: Thanks!

davidemyers:
FYI to anyone who has tried or would like to try Quad9 while in beta, they changed the names of their servers and I wouldn't be surprised if they change them again. Changes like this unfortunately
break DNS.
```
dnscrypt_servers:
  ipv4:
#   - cloudflare
#   - google
    - dnscrypt-ip4-filter-pri
  ipv6:
#   - cloudflare-ipv6
    - dnscrypt-ip6-filter-pri
```
On an already deployed server, edit `/etc/dnscrypt-proxy/dnscrypt-proxy.toml`:
```
server_names = ['dnscrypt-ip4-filter-pri', 'dnscrypt-ip6-filter-pri']
```

dguido:
Thanks!

Issue #1145, opened 2018-10-09
Ubunt 18.04 + ALGO ERROR . The python pyOpenSSL module is required.


plbartek89:
Hello , 
I try to install algo on my ubuntu 18.04 VPS  but no success atm.
[https://gyazo.com/67f8fe27660b3b827b37f58e3f66191d](url)

`TASK [vpn : Create links for the private keys] *******************************************************************************
failed: [localhost -> localhost] (item=daxcvxn) => {"changed": false, "item": "daxcvxn", "msg": "the python pyOpenSSL module i                      s required"}
failed: [localhost -> localhost] (item=jaxcvxvck) => {"changed": false, "item": "jaxcvxvck", "msg": "the python pyOpenSSL modu                      le is required"}
`

davidemyers:
I was able to reproduce this error when attempting a local install on a DO Droplet.

yilmzfurkan:
Reproduced the same. Manually installed pyOpenSSL and made a clean installation everything worked. 

`pip install pyopenssl`

plbartek89:
@yilmzfurkan
Thanks man.
Works like a charm.
I have one more question about speed my vps has CPU: 2 vCores
RAM: 1GB  Bandwidth: 200Mbps my internet connection 250Mbps. When i do speedtest with vpn on I reach +/- 20Mbps only. What can i do to increase the speed a little bit , when possible ofc. ?

davidemyers:
@yilmzfurkan That command works for me too, when run outside of Algo's Python virtual environment.

@plbartek89 This might have nothing to do with your speed issue, but one of the DigitalOcean Droplets I created in NYC3 today is crazy slow, while others in NYC3 and SFO2 are fine. Maybe you've got a slow VPS.

davidemyers:
I think the local install instructions as currently written are obsolete. The normal instructions appear to be sufficient to do a local install on Ubuntu 18.04, with the exception of this issue.

As for this issue, I'm not sure if Jack would consider it a bug or something to be handled through documentation.

Issue #1139, opened 2018-10-04
What ports need to be left as open for Algo to function properly


AlexShapka:
Unfortunately, I was not able to add more comments on the closed https://github.com/trailofbits/algo/issues/1136, so I hope it's ok to open a new issue, considering the subject matter was not properly addressed. 

Jack, thanks for the update, however:

1. I know how to change the ports, but the problem is that as I described as soon as I change the VPN falls. There is some kind of Algo-related configuration somewhere which sets port 22 to use and I need to get it changed. So please do not mark this issue as closed unless the issue has been resolved.

2. I travel a lot, so I have dynamic address. I do appreciate your advice, but would be really nice to concentrate on the asked question.

AlexShapka:
The reported issue could also caused by my attempt to leave only the custom port open closing all other ports, including port 22.  So maybe it is not at all about the port where SSH deamon listens to. I wonder does Algo need some particular ports to be left open?

jackivanov:
We use 22 ports for SSH only (and ssh tunnels if enabled) and neither IPsec nor WireGuard don't depend on this port.
The only ports you need to get IPsec working are 500 and 4500 UDP and 51820 for WireGuard

AlexShapka:
Thanks for the update, however it didn't work. Even after digging into this little bit further and finding out I needed to also leave ports 50 and 51 open for TCP, unfortunately, as soon as the firewall starts the Internet is gone. Surprisingly, SSH connection to the remote VPN server is intact, so it is not about SSH port anymore. I am afraid some more ports need to be left as open.

Just to be sure I've included all involved ports both in TCP and UDP setting of my APF firewall like so:

```
# Common inbound (ingress) TCP ports
IG_TCP_CPORTS="22,50,51,500,4500,51820"

# Common inbound (ingress) UDP ports
IG_UDP_CPORTS="50,51,500,4500,51820"
```

with no success. 

I just wonder didn't anyone before hardened the remote Ubuntu server with Algo client running in it to know the exact ports?

AlexShapka:
I've changed the title as it is not about SSH port 22 anymore. SSH works just fine both on the default port 22 or any other custom port. However, as soon as APF starts effectively closing all the ports except mentioned in its TCP and UDP configuration settings the whole things stops working - SSH connection to the Algo client is still intact, but Internet is not available through browser or pinging any other website via CLI.

jackivanov:
[...]

Issue #1133, opened 2018-09-29
error when deploy on Ubuntu cloud host


shi-xiaopeng:
I want to deploy algovpn on my Ali-cloud host which using Ubuntu 16.04 64bit.
I follow the README deploy instruction well before running ./algo , and then it come to an exception
````
TASK [wireguard : Generate private keys] *********************************************************************************************************************
 [WARNING]: Unable to find 'configs/47.**.**.138/wireguard//private/shi' in expected paths (use -vvvvv to see paths)


TASK [wireguard : Generate public keys] **********************************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'file'. Error was a <class 'ansible.errors.AnsibleError'>, original message: could not locate file in lookup: configs/47.****.138/wireguard//private/shi"}
 ````
I search it on web, it seems  all stuff about Ansible. What can i do, any hint ?

The full log is below:

```
(env) root@shi:~/algo# ./algo

PLAY [Ask user for the input] ********************************************************************************************************************************

TASK [Gathering Facts] ***************************************************************************************************************************************
ok: [localhost]
[pause]
What provider would you like to use?
    1. DigitalOcean
    2. Amazon EC2
    3. Vultr
    4. Microsoft Azure
    5. Google Compute Engine
    6. Scaleway
    7. OpenStack (DreamCompute optimised)
    8. Install to existing Ubuntu 18.04 server (Advanced)
  
Enter the number of your desired provider
:
8

TASK [pause] *************************************************************************************************************************************************
ok: [localhost]

TASK [Set facts based on the input] **************************************************************************************************************************
ok: [localhost]
[pause]
Do you want macOS/iOS clients to enable "VPN On Demand" when connected to cellular networks?
[y/N]
:
y

TASK [pause] *************************************************************************************************************************************************
ok: [localhost]
[pause]
Do you want macOS/iOS clients to enable "VPN On Demand" when connected to Wi-Fi?
[y/N]
:
y

TASK [pause] *************************************************************************************************************************************************
ok: [localhost]
[pause]
List the names of trusted Wi-Fi networks (if any) that macOS/iOS clients exclude from using the VPN
(e.g., your home network. Comma-separated value, e.g., HomeNet,OfficeWifi,AlgoWiFi)
:


TASK [pause] *************************************************************************************************************************************************
ok: [localhost]
[pause]
Do you want to install a DNS resolver on this VPN server, to block ads while surfing?
[...]

Issue #1119, opened 2018-09-18
"Service is in unknown state"


broady:
**Describe the bug**

Running Ubuntu 18.04 on a Raspberry Pi v3, `./algo` fails on the "systemd services enabled and started" step.

**To Reproduce**

1. Install Ubuntu 16.04 on Raspberry Pi.
2. Update to 18.04.
2. Install algo.
3. Run `sudo ./algo`.

**Full log**

```
$ sudo ./algo

PLAY [Ask user for the input] ************************************************************************************************************************


TASK [Gathering Facts] *******************************************************************************************************************************
ok: [localhost]
[pause]
What provider would you like to use?
    1. DigitalOcean
    2. Amazon EC2
    3. Vultr
    4. Microsoft Azure
    5. Google Compute Engine
    6. Scaleway
    7. OpenStack (DreamCompute optimised)
    8. Install to existing Ubuntu 18.04 server (Advanced)

Enter the number of your desired provider
:
8

TASK [pause] *****************************************************************************************************************************************
ok: [localhost]

TASK [Set facts based on the input] ******************************************************************************************************************
ok: [localhost]
[pause]
Do you want macOS/iOS clients to enable "VPN On Demand" when connected to cellular networks?
[y/N]
:


TASK [pause] *****************************************************************************************************************************************
ok: [localhost]
[pause]
Do you want macOS/iOS clients to enable "VPN On Demand" when connected to Wi-Fi?
[y/N]
:


TASK [pause] *****************************************************************************************************************************************
ok: [localhost]
[pause]
Do you want to install a DNS resolver on this VPN server, to block ads while surfing?
[y/N]
:


TASK [pause] *****************************************************************************************************************************************
ok: [localhost]
[pause]
Do you want each user to have their own account for SSH tunneling?
[y/N]
:


TASK [pause] *****************************************************************************************************************************************
ok: [localhost]
[pause]
Do you want the VPN to support Windows 10 or Linux Desktop clients? (enables compatible ciphers and key exchange, less secure)
[y/N]
:


TASK [pause] *****************************************************************************************************************************************
ok: [localhost]
[pause]
Do you want to retain the CA key? (required to add users in the future, but less secure)
[y/N]
:


TASK [pause] *****************************************************************************************************************************************
ok: [localhost]

TASK [Set facts based on the input] ******************************************************************************************************************
ok: [localhost]

PLAY [Provision the server] **************************************************************************************************************************

[...]

PR #1148, opened 2018-10-12
Update local install instructions

davidemyers:
## Description

Simplified the local installation instructions.

## Motivation and Context

The instructions appeared to be out-of-date, as mentioned in #1145. No special steps appear to be necessary to do a local install.

## How Has This Been Tested?

Deployed to a local server using the changes in #1147. Others might wish to test this as well.

Did not test the option to skip `iptables`, just carried the instructions forward.

## Types of changes

## Checklist:

### DIFF
--- a/docs/deploy-to-ubuntu.md
+++ b/docs/deploy-to-ubuntu.md
@@ -1,6 +1,6 @@
 # Local deployment
 
-You can use Algo to configure an existing server as an Algo VPN rather than have Algo create and configure a new server on a cloud provider.
+You can use Algo to configure a local server as an Algo VPN rather than create and configure a new server on a cloud provider.
 
 Install the Algo scripts on your server and follow the normal installation instructions, then choose:
 ```
@@ -8,4 +8,4 @@ Install to existing Ubuntu 18.04 server (Advanced)
 ```
 Make sure your server is running the operating system specified.
 
-**PLEASE NOTE**: Algo is intended to be used to create a dedicated VPN server. If you install Algo on an existing server any existing services might break. In particular, the firewall rules will be overwritten. If you don't want to overwrite the rules you must deploy via `ansible-playbook` and skip the `iptables` tag as described in [deploy-from-ansible.md](deploy-from-ansible.md), after which you'll need to implement the necessary rules yourself.
+**PLEASE NOTE**: Algo is intended for use as a _dedicated_ VPN server. If you install Algo on an existing server, then any existing services might break. In particular, the firewall rules will be overwritten. If you don't want to overwrite the rules you must deploy via `ansible-playbook` and skip the `iptables` tag as described in [deploy-from-ansible.md](deploy-from-ansible.md), after which you'll need to implement the necessary rules yourself.
--- a/docs/deploy-to-ubuntu.md
+++ b/docs/deploy-to-ubuntu.md
@@ -1,20 +1,11 @@
 # Local deployment
 
-It is possible to download the Algo scripts to your own Ubuntu 18.04 server and run the scripts locally.
+You can use Algo to configure an existing server as an Algo VPN rather than have Algo create and configure a new server on a cloud provider.
 
-In order to start, you need to install Ansible. Installing Ansible via pip requires pulling in a lot of dependencies, including a full compiler suite. It would be easier to use apt, however, Ubuntu 18.04 only comes with Ansible 2.0.0.2. The easiest solution is to install the Ansible PPA for a newer version of Ansible via apt, however, using a PPA requires installing `software-properties-common`.
-
-tl;dr:
-
-```shell
-sudo apt-get install software-properties-common && sudo apt-add-repository ppa:ansible/ansible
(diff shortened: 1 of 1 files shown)


TC1977:
@davidemyers _Deleting_ the local install instructions wasn't what I had in mind. 😅 

Using the usual install instructions (i.e., 
```
git clone https://github.com/trailofbits/algo		
cd algo
sudo apt-get update && sudo apt-get install \
    build-essential \
    libssl-dev \
    libffi-dev \
    python-dev \
    python-pip \
    python-setuptools \
    python-virtualenv -y
python -m virtualenv --python=`which python2` env &&
    source env/bin/activate &&
    python -m pip install -U pip &&
    python -m pip install -r requirements.txt
```
on an AWS Ubuntu 18.04 instance fails with the following:

```ssh

TASK [vpn : Create links for the private keys] ***************************************************************************
failed: [localhost -> localhost] (item=dan) => {"changed": false, "item": "dan", "msg": "the python pyOpenSSL module is required"}
failed: [localhost -> localhost] (item=jack) => {"changed": false, "item": "jack", "msg": "the python pyOpenSSL module is required"}

```

TC1977:
Okay, here's something weird. The iptables rules weren't overwritten.

```ssh
ubuntu@ip-172-31-35-19:~/algo$ sudo iptables -S
-P INPUT ACCEPT
-P FORWARD ACCEPT
-P OUTPUT ACCEPT
```

TC1977:
> My thinking was that duplicating the instructions will just lead to maintenance headaches when both sets have to be updated in lockstep.

How about at the very least, just explicitly instructing `git clone https://github.com/trailofbits/algo` and `cd algo`, then a [link back to the readme](https://github.com/trailofbits/algo/commits/master/README.md) for the base instructions?

> Also, the local install prompt says "Advanced", which I interpret as meaning inexperienced users should probably not try it.

True, but when has that stopped anyone? And even experienced users appreciate cut-and-paste. 😄

TC1977:
Doesn't have to be `git clone https://github.com/trailofbits/algo`, I suppose, but _some_ explicit instructions we can simply cut and paste would be nice. Every other page in the docs has them.

davidemyers:
Can Dan or Jack cast a tie-breaking vote? :smile:

I don't mind duplicating the instructions here, my concern is that will make this page more likely to become incorrect whenever the instructions in the README are updated, like it is now.

davidemyers:
Here's what I would add:

----
Here is an Ubuntu-specific summary of the commands to run to perform a local deployment. For a local deployment, Algo is installed and run on the Ubuntu Server instance that is to become a VPN server, whereas with a cloud deployment Algo is installed and run on a separate system. Also, for a local deployment you'll probably want to run all of these commands as `root`.

Install the prerequisite system packages:
```shell
sudo apt update && sudo apt install \
    build-essential \
    libssl-dev \
    libffi-dev \
    python-dev \
    python-pip \
    python-setuptools \
    python-virtualenv -y
```
Since `unzip` is not installed by default it's easier to clone the GitHub repository than download the ZIP file:
```shell
git clone https://github.com/trailofbits/algo 
cd algo
```
Install the remaining packages in a Python virtual environment and activate the environment:
```shell
python -m virtualenv --python=`which python2` env &&
    source env/bin/activate &&
    python -m pip install -U pip &&
    python -m pip install -r requirements.txt
```
Edit the file `config.cfg` with `vi` or `nano` to add users and then run Algo:
```shell
vi config.cfg
./algo
```

jackivanov:
I don't think we need this page at all (deploy-to-ubuntu.md). There are instruction in the main readme already. And as David said, duplicating the instructions will just lead to maintenance headaches

davidemyers:
I think a warning that a local install assumes a dedicated system (and will stomp on existing firewall rules) will head off some user issues, but it doesn't have to be on this page.

TC1977:
@jackivanov The first line of the main readme instructions ("The easiest way to get an Algo server running is to let it set up a new virtual machine in the cloud for you. ") implies a cloud install. [Cloud install from Windows](https://github.com/trailofbits/algo/blob/master/docs/deploy-from-windows.md) and [local deploy to FreeBSD](https://github.com/trailofbits/algo/blob/master/docs/deploy-to-freebsd.md) each get their own page. The local deploy to Ubuntu option seems to have more people using it, or at least getting more problems with it, judging by the number of issues filed (#1145, #1139, #1133, #1119, #1118 just from the last month). Deleting the local deploy page entirely won't keep people from running into problems, unless you promote the instructions to the main readme page (where it probably doesn't belong). 

If duplicating the explicit instructions leads to headaches, you could always just link back to the main readme.

dguido:
I guess all the virtualenv stuff and recent refactoring eliminates the need for this page? I think I'm with Jack, let's just deprecate this page entirely if it's not needed.

davidemyers:
I think leaving the page could be helpful because there seem to be recurring issues where the user wants to use their server for multiple purposes and then goes and messes up the firewall. There are also users who want to use their server for local network ingress and this page could either contain tips for doing that or could discourage the practice altogether.

dguido:
Ok we'll leave it for now.

PR #1183, opened 2018-10-31
WireGuard update-users fix

### LINKED ISSUE #1131: New Wireguard configs required after ./algo update-users
**Describe the bug**
I don't have an Android phone, or a Wireguard client for that matter, and can't reproduce this.

Gitter.im report:

Krzysztof Szromek @szromek Sep 26 06:21
Hi guys! I've run into one issue - every time I update list of users, all Android clients stop working and they need to upload new Wireguard configs which is really hideous process. Is this something we can avoid?

**To Reproduce**

1. Install Algo with Wireguard clients on Android phones
2. Run ./algo update-users (I'd assume with `keys_clean_all` flag in `config.cfg` set to False)
3. Existing Wireguard clients won't work - have to download new config files

**Expected behavior**

Should be able to continue using existing Wireguard clients, without having to download new config files (unless of course the associated user was removed)

jackivanov:
## Description
Due to the WireGuard implementation we have to define IP addresses for each user explicitly. Previously, it was done in sequential order, with IPs stored in memory, and it was the root cause of [this](1131) bug. This PR brings ability not to restrict users by IP addresses, I don't think it makes any sense n this implementation.

## Motivation and Context
Fixes #1131

## How Has This Been Tested?
Deployed an algo server

## Types of changes

## Checklist:

### DIFF
--- a/roles/vpn/defaults/main.yml
+++ b/roles/vpn/defaults/main.yml
@@ -7,13 +7,13 @@ wireguard_network_ipv4:
   prefix: 24
   gateway: 10.19.49.1
   clients_range: 10.19.49
-  clients_start: 100
+  clients_start: 2
 wireguard_network_ipv6:
   subnet: 'fd9d:bc11:4021::'
   prefix: 48
   gateway: 'fd9d:bc11:4021::1'
   clients_range: 'fd9d:bc11:4021::'
-  clients_start: 100
+  clients_start: 2
 wireguard_vpn_network: "{{ wireguard_network_ipv4['subnet'] }}/{{ wireguard_network_ipv4['prefix'] }}"
 wireguard_vpn_network_ipv6: "{{ wireguard_network_ipv6['subnet'] }}/{{ wireguard_network_ipv6['prefix'] }}"
 keys_clean_all: false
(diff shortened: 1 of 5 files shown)


davidemyers:
I think there still might be issues with this approach.

If I deploy using the default user list, `dan` gets address `.101` and `jack` gets `.102` in their client config files.

If I add `alice` as the first user in `config.cfg` then run `./algo update-users`, `alice` gets `.101`, `dan` gets `.102`, and `jack` gets `.103`.

So unless all users replace their configs after `update-users`, `alice` and `dan` will both try to use the same address (which I haven't tested but don't think would work properly).

digeratus:
How can I help move this passed WIP? Is testing necessary or still a bug?

jackivanov:
I forgot to publish the changes. Check it out now, please

davidemyers:
This looks good to me.

I ran the same test as above. After `update-users`, `dan.conf` and `jack.conf` were unchanged and `alice.conf` received the next consecutive IP address. I was able to connect properly with `alice.conf`.

I have a question about the IPv6 addresses, though. The `dan.conf` file contains:
```
Address = 10.19.49.3/32,fd9d:bc11:4021::101/48
```
The same values are used on the server for `AllowedIPs`. Should the IPv6 address start at the same index as the IPv4 address, and should the prefix be /128?

davidemyers:
Now I get on the client:
```
Address = 10.19.49.3/24,fd9d:bc11:4021::3/48
```
And on the server:
```
AllowedIPs = 10.19.49.3/32,fd9d:bc11:4021::3/128
```
These look correct to me and I can connect OK.

jackivanov:
@davidemyers Yes, these are the correct setting for client and server

digeratus:
Can someone please merge this?

digeratus:
Thank you 🙏

Issue #1221, opened 2018-11-27
DNS stops working with ad blocking installed


davidemyers:
Some users have reported issues with DNS not working either immediately upon deployment or after a few days. Some have reported that the problem does not occur if they deploy with `dns_encryption: false`.

@kristallizer [discovered](https://github.com/trailofbits/algo/issues/1130#issuecomment-441827271) in #1130 that `dnsmasq` failed to restart after `/usr/local/sbin/adblock.sh` had run. The failure appears to be during parsing of config files with `dnsmasq --test`:
```
root@algo-ny-2-encr:~# service dnsmasq status
● dnsmasq.service - dnsmasq - A lightweight DHCP and caching DNS server
   Loaded: loaded (/lib/systemd/system/dnsmasq.service; enabled; vendor preset: enabled)
  Drop-In: /etc/systemd/system/dnsmasq.service.d
           └─100-CustomLimitations.conf
   Active: failed (Result: signal) since Mon 2018-11-26 02:10:43 UTC; 20h ago
  Process: 25008 ExecStop=/etc/init.d/dnsmasq systemd-stop-resolvconf (code=exited, status=0/SUCCESS)
  Process: 20173 ExecStartPost=/etc/init.d/dnsmasq systemd-start-resolvconf (code=exited, status=0/SUCCESS)
  Process: 20152 ExecStart=/etc/init.d/dnsmasq systemd-exec (code=exited, status=0/SUCCESS)
  Process: 25015 ExecStartPre=/usr/sbin/dnsmasq --test (code=killed, signal=KILL)
 Main PID: 20170 (code=exited, status=0/SUCCESS)
      CPU: 210ms

Nov 26 02:10:39 algo-ny-2-encr systemd[1]: Starting dnsmasq - A lightweight DHCP and caching DNS server...
Nov 26 02:10:43 algo-ny-2-encr systemd[1]: dnsmasq.service: Control process exited, code=killed status=9
Nov 26 02:10:43 algo-ny-2-encr systemd[1]: dnsmasq.service: Failed with result 'signal'.
Nov 26 02:10:43 algo-ny-2-encr systemd[1]: Failed to start dnsmasq - A lightweight DHCP and caching DNS server.
Nov 26 02:10:43 algo-ny-2-encr systemd[1]: dnsmasq.service: Consumed 210ms CPU time
```

[...]

PR #1227, opened 2018-11-30
Run adblock.sh at a random time

davidemyers:
## Description

For installations that use ad blocking, schedule the running of `adblock.sh` at a different, randomly chosen time for each installation.

## Motivation and Context

As noted in #1221 there's evidence that the daily run of `adblock.sh` at 02:10 UTC is overwhelming one or more of the hosts that serve the block lists.

There may still be weaknesses in `adblock.sh` that can cause `dnsmasq` to fail. This change hopefully makes any problems less likely to occur.

## How Has This Been Tested?

Deployed three times to DigitalOcean and checked the `cron` jobs. Did not test with BSD but since the change is to Ansible there should be no issues.

## Types of changes

## Checklist:

### DIFF
--- a/roles/dns_adblocking/tasks/main.yml
+++ b/roles/dns_adblocking/tasks/main.yml
@@ -30,8 +30,8 @@
     - name: Adblock script added to cron
       cron:
         name: Adblock hosts update
-        minute: 10
-        hour: 2
+        minute: "{{ range(0, 60) | random }}"
+        hour: "{{ range(0, 24) | random }}"
         job: /usr/local/sbin/adblock.sh
         user: root
 


TC1977:
While it would certainly spread out the load on the adblock list server over the day, randomizing both the hour and second has the potential of introducing a couple seconds of DNS downtime at a pretty inconvenient time, every day.

Would it be better to leave the hour at 2 and only randomize the minute? That would probably be enough to decrease the load on whichever server we're talking about, and at least not worsen the downtime issue (I recognize that the whole world doesn't live on UTC, or UTC-4).

davidemyers:
I wasn't sure how to choose a time that would work best for every geography so I made it "fair", but it would be easy enough to limit the range of hours chosen if there's agreement on what that range should be.

davidemyers:
I don't think Ansible knows where a server is geographically located.

And wouldn't it be most convenient if maintenance tasks like this were scheduled with respect to _client_ local time?

So it's complicated.

Personally, I set the time zone on the servers I deploy, but we can't expect most users to do that.

TC1977:
Right. Anyone who really cares about two seconds of downtime out of 86400 is probably also going to set their server time zone. And anyone who can actually find the dnsmasq reload time as the source of their two-second downtime has probably already set it. 😄

TC1977:
```ssh

Algo running on: Ubuntu 18.04.1 LTS (Virtualized: xen)
Created from git fork. Last commit: a6cf2ce Run adblock.sh at a random time
Python 2.7.15rc1
Runtime variables:
    algo_provider "local"
    algo_ondemand_cellular "False"
    algo_ondemand_wifi "False"
    algo_ondemand_wifi_exclude "_null"
    algo_local_dns "True"
    algo_ssh_tunneling "False"
    algo_windows "False"
    wireguard_enabled "True"
    dns_encryption "True"
```
```ssh
ubuntu@ip-172-16-254-54:/etc$ sudo crontab -u root -l
#Ansible: Adblock hosts update
33 23 * * * /usr/local/sbin/adblock.sh
```

jackivanov:
> 02:10 UTC

It's wrong. Nothing to do with Ansible either. Cron runs this job at 2:10 server local time, not UTC

jackivanov:
I'm not sure if it can be addressed in this PR, but here is a project I found - https://github.com/notracking/hosts-blocklists. We may just replace our lists. githubusercontent would be more reliable than unknown sites.

davidemyers:
The list servers certainly appear to be getting clobbered by Algo at 02:10 UTC. Here are logs from two different instances in DO NYC3, the first with the default time and the second randomized:
```
Nov 30 02:10:27 vpn8 dnsmasq[23755]: using 5991 more local addresses
Dec 01 02:10:33 vpn8 dnsmasq[24887]: using 13920 more local addresses
Dec 02 02:10:21 vpn8 dnsmasq[26336]: using 13920 more local addresses
Dec 03 02:10:22 vpn8 dnsmasq[26995]: using 12811 more local addresses
```
```
Nov 30 19:38:09 vpn9 dnsmasq[17071]: using 56274 more local addresses
Dec 01 19:38:09 vpn9 dnsmasq[18012]: using 56272 more local addresses
Dec 02 19:38:10 vpn9 dnsmasq[19043]: using 56272 more local addresses
```
If randomization is an acceptable approach, and they way I implemented it is correct, can this PR be merged while other solutions are explored?

TC1977:
> > 02:10 UTC
> 
> It's wrong. Nothing to do with Ansible either. Cron runs this job at 2:10 server local time, not UTC.
> The server timezone should be set according to the region the server running in and it mostly depends on how the provider configures server images. I suppose, all the providers we support do it right.

At least on AWS instances, server local time *is* UTC. There might be an environment variable I don't know about (I don't know a lot of things), and other cloud providers might have such a variable, but I couldn't find any way of obtaining the server local time from the region name, say, and setting it through ansible. I suppose you could have the user input the desired time zone in the beginning and then have ansible set the clock.

Only other issue I see is this PR won't affect existing installations, so the problem will still remain until the installed user base upgrades. No good way to correct that, though.

jackivanov:
Yeah, I was wrong, the providers don't modify timezones according to the regions, oh crap. Yes, let's merge it.
So, let's merge it

PR #1253, opened 2018-12-17
Replace 'max_mss' with 'reduce_mtu'

### LINKED ISSUE #1217: For GCE deployments, automatically set max_mss, or prompt the user to set it
Prior to #976, `max_mss` used to be set for GCE deployments automatically (see https://github.com/trailofbits/algo/pull/976/files#diff-28f2b95533afb47cbec1d823b0f1a941L535). Now that `max_mss` is undefined, GCE deployments no longer appear to work out of the box. I brought up a new GCE deployment on November 23, and noticed I was unable to load sites in my browser until I redeployed with `max_mss` set to 1316, the default value listed in **config.cfg**.

Can the previous behavior of automatically setting `max_mss` for GCE deployments be brought back, or can the user be prompted to set `max_mss` (maybe default to 1316)?

If either approach sounds good, I can open a PR with the changes.

Thanks!

davidemyers:
Replaces the `max_mss` configuration variable with a new variable `reduce_mtu`  used to calculate both the MSS clamping values for IPsec and the MTU value for WireGuard. Also sets it automatically based on the server MTU.

## Description

Previously, MSS clamping on the server was used for both IPsec and WireGuard connections, even though WireGuard can set the MTU directly in its config files and has less tunnel overhead than IPsec. This change makes use of WireGuard's ability to set the MTU on the client side. The optimal MTU settings for IPsec and WireGuard are different and this allows them to be adjusted together by changing a single variable.

This change also automatically reduces the MTU if the MTU on the server is found to be less than 1500, but only if the user leaves `reduce_mtu` at the default of `0`. Otherwise the user value is used.

Apple devices use an MTU of 1400 for IPsec, and the Linux and iOS WireGuard implementations use an MTU of 1420. Therefore the MTU and MSS values used when `reduce_mtu > 0` are:

WireGuard MTU = 1420 - `reduce_mtu`
IPsec IPv4 MSS = 1400 - 40 - `reduce_mtu`
IPsec IPv6 MSS = 1400 - 60 - `reduce_mtu`

## Motivation and Context

The user should not have to understand the concept of MSS.

The user should not have to change any settings in order to deploy to providers with smaller MTUs like GCE.

Closes #1217.

## How Has This Been Tested?

Deployed to DigitalOcean to verify the config file changes. Deployed to GCE to verify no hang when accessing Google properties over IPsec and WireGuard. 

More testing should be done and the rewritten troubleshooting instructions should be reviewed.

## Types of changes

## Checklist:

### DIFF
--- a/config.cfg
+++ b/config.cfg
@@ -21,15 +21,14 @@ vpn_network_ipv6: 'fd9d:bc11:4020::/48'
 wireguard_enabled: true
 wireguard_port: 51820
 
-# MSS is the TCP Max Segment Size
-# Setting the 'max_mss' Ansible variable can solve some issues related to packet fragmentation
-# This appears to be necessary on (at least) Google Cloud,
-# however, some routers also require a change to this parameter
-# See also:
-# - https://github.com/trailofbits/algo/issues/216
-# - https://github.com/trailofbits/algo/issues?utf8=%E2%9C%93&q=is%3Aissue%20mtu
-# - https://serverfault.com/questions/601143/ssh-not-working-over-ipsec-tunnel-strongswan
-#max_mss: 1316
+# Reduce the MTU of the VPN tunnel
+# Some cloud and internet providers use a smaller MTU (Maximum Transmission
+# Unit) than the normal value of 1500 and if you don't reduce the MTU of your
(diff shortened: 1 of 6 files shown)


jackivanov:
Looks good to me

PR #1269, opened 2018-12-23
Install resolvconf on Ubuntu to fix DNS configurations

faf0:
The local (non-cloud) installation process changes /etc/resolv.conf such that
127.0.0.53 is the only nameserver. Moreover, the dnscrypt-proxy
configuration relies on 127.0.0.53:53 as the fallback DNS server.
However, there is no DNS server listening on that IP unless the package
resolvconf is installed.

Thus, a local installation of algo on Ubuntu breaks the DNS
configuration. The issue is fixed by installing resolvconf.

### DIFF
--- a/roles/common/tasks/ubuntu.yml
+++ b/roles/common/tasks/ubuntu.yml
@@ -7,6 +7,7 @@
     with_items:
       - python2.7
       - sudo
+      - resolvconf
 
   - name: Ubuntu | Configure defaults
     alternatives:


davidemyers:
On a normal Ubuntu 18.04 system, `systemd-resolved` listens on 127.0.0.53 and is configured from `netplan`. See the [Ubuntu Server Guide](https://help.ubuntu.com/lts/serverguide/network-configuration.html.en#name-resolution). In a typical installation `systemd-resolved` will use DNS servers obtained via DHCP. Running `systemd-resolve --status` shows what DNS servers are being used.

None of that has anything to do with Algo and I'm pretty sure Algo doesn't change how the server uses DNS.

davidemyers:
I'm unable to recreate this issue when performing local installs on servers at DigitalOcean. I'm not sure what's different about your server.

What output do you get from the command `ip addr show dev lo`? On a working system I see four addresses:
```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet 172.16.0.1/32 scope global lo
       valid_lft forever preferred_lft forever
    inet6 fcaa::1/64 scope global 
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
```

faf0:
After a failed install, I get the same output as you.

Before I install Algo, I just get the first and last interfaces (i.e., `127.0.0.1/8` and `::1/128`).

FWIW, when I set `local_service_ip: 127.0.0.53` in `config.cfg` on a fresh machine, the installation succeeds.

davidemyers:
Have you had any luck figuring out the cause of your issue?

Do you have any 172.* addresses on any of your other network interfaces? I'm not certain that would be a problem but it's worth checking.

You might want to close this PR if the root cause of the issue is still unknown.

PR #1281, opened 2018-12-30
Support for associating to existing AWS Elastic IP

statik:
## Description

This adds the option of associating the Algo instance to an existing Elastic IP in AWS instead of provisioning a new Elastic IP.

## Motivation and Context
I have provisioned long-lived Elastic IPs with reverse DNS configured (https://aws.amazon.com/blogs/aws/reverse-dns-for-ec2s-elastic-ip-addresses/) and whitelists configured for those IPs at customer sites. I would like the ability to use one of my existing Elastic IPs when provisioning an Algo instance instead of provisioning a brand new EIP.

## How Has This Been Tested?
I have provisioned new algo instances with and without associating to pre-existing Elastic IP, and confirmed that the cloudformation stack deletes cleanly.

## Types of changes

- [] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [x] I have read the **CONTRIBUTING** document.
- [x] My code follows the code style of this project.
- [] My change requires a change to the documentation.
- [] I have updated the documentation accordingly.
- [] I have added tests to cover my changes.
- [] All new and existing tests passed.

PR #1292, opened 2019-01-08
AWS support for existing EIP (revised)

statik:
## Description

This adds the option of associating the Algo instance to an existing Elastic IP in AWS instead of provisioning a new Elastic IP.

## Motivation and Context
I have provisioned long-lived Elastic IPs with reverse DNS configured (https://aws.amazon.com/blogs/aws/reverse-dns-for-ec2s-elastic-ip-addresses/) and whitelists configured for those IPs at customer sites. I would like the ability to use one of my existing Elastic IPs when provisioning an Algo instance instead of provisioning a brand new EIP.

## How Has This Been Tested?
This is a followup to #1281. I removed the dependency on Ansible 2.6 following the example of the local lightsail facts module and tested the changes work with a fresh virtualenv.

I have provisioned new algo instances with and without associating to pre-existing Elastic IP, and confirmed that the cloudformation stack deletes cleanly.

## Types of changes

## Checklist:


jackivanov:
We need to get this feature documented also

roles/cloud-ec2/tasks/prompts.yml
```diff
@@ -53,3 +53,28 @@
...
+          {% endfor %}
+          ['']
+    register: _use_existing_eip
+
+  - set_fact:
+      use_existing_eip: >-
+        {% if _use_existing_eip.user_input is defined and _algo_region.user_input != "" %}
+        {{ available_eip_addresses[_use_existing_eip.user_input | int -1 ]['allocation_id'] }}
+        {%- else %}{% endif %}
+  when: use_existing_eip is undefined
```
jackivanov: Can we run this block only when `use_existing_eip` is set to true in config.cfg? By default it should be false

statik: Sure, that makes a lot of sense! I just pushed a commit changing this.

jackivanov: Thanks. I guess the installation will fail, because there is no default for this variable, so you also need to add `use_existing_eip: false` to config.cfg, and add a description. Also, I see that you use the same variable in `set_fact`, that probably needs to be renamed to get it more understandable

statik: Thanks! I want to cooperate, but I guess I don't understand how you want the behavior to change. Are you also wanting the prompt asking if you want to use an existing EIP to be hidden if use_existing_eip is false in config.cfg?

jackivanov: No, I want `use_existing_eip` to be set to `false` in config.cfg, and the prompt asking only if it's set to `true`

jackivanov:
Ping @statik Would you be able to implement the changes and fix the conflicts?

statik:
Thank you @jackivanov! I have been flooded with some personal responsibilities and got stuck on this, thank you for pushing it forward and all your work on Algo.

PR #1293, opened 2019-01-09
Add note about new WireGuard for iOS default MTU

davidemyers:
## Description

Update the troubleshooting section on MTU issues to point out that the default MTU for WireGuard for iOS is now 1280 (was 1420) but that Algo might have changed it.

## Motivation and Context

If Algo detects the server has a smaller than normal MTU, the generated WireGuard client config files will contain an MTU directive intended to match the client MTU with that of WireGuard on the server. This should normally be the right thing to do, but now in the case of WireGuard for iOS this will generally cause the MTU to be larger than the new default of 1280. The user might want to delete the MTU directive added by Algo in order to use the new default.

## How Has This Been Tested?

No testing, documentation.

## Types of changes

## Checklist:

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -391,6 +391,10 @@ sudo ifconfig eth0 mtu 1440
 ```
 To make the change take affect after a reboot, on Ubuntu 18.04 and later edit the relevant file in the `/etc/netplan` directory (see `man netplan`).
 
+#### Note for WireGuard iOS users
+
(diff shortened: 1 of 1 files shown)


davidemyers:
I added setting the MTU in the WireGuard configs as part of the `reduce_mtu` PR. It gets added only if the MTU on the server is detected to be less than 1500, such as when deploying to GCE. Without it GCE users would see hanging connections when trying to access Google sites. It's still needed for non-iOS WireGuard clients, including those who will use the forthcoming macOS app, so this is my attempt to address the situation through documentation. I'm not sure what the Android app does, though.

I think the root of problem is that Algo generates one-size-fits-all WireGuard config files, but it's starting to look like one size does not fit all. Mobile clients probably should not be using `PersistentKeepalive`, which Algo always adds (see a recent comment by Jason [here](https://github.com/trailofbits/algo/commit/dbd68aa97d81e3286264344c112744554e2bd5b3#r31803542)), and as I've documented elsewhere I think the `DNS` directive should be changed for Ubuntu clients.

PR #1345, opened 2019-03-05
Apache CloudStack support

0xmilkmix:
This new role allows to provision instances on Apache CloudStack providers like Exoscale.

I initially wanted to add a role specific for Exoscale but since they are using Apache CloudStack, better make it generic :)

## Description
I added a new role `cloud-cloudstack` and modified the required files for inputs and configurations defaults.

## Motivation and Context
I got an Exoscale gift-card to try their service at the latest Lausanne DevSecOps meetup. Since they advertised the non-US feature of their business, we discussed VPN and the like. What better use than add support for their cloud for this gift card?

## How Has This Been Tested?
I only tested this CloudStack role on Exoscale. However, the only things to change would be in the user's `$HOME/.cloudstack.ini` file to specify the endpoint to use.

## Types of changes

## Checklist:

### DIFF
--- a/config.cfg
+++ b/config.cfg
@@ -150,6 +150,10 @@ cloud_providers:
   openstack:
     flavor_ram: ">=512"
     image:  Ubuntu-18.04
+  cloudstack:
+    size: Tiny
+    image: Linux Ubuntu 18.04 LTS 64-bit
+    disk: 10
   vultr:
     os: Ubuntu 18.04 x64
     size: 1024 MB RAM,25 GB SSD,1.00 TB BW
--- a/config.cfg
+++ b/config.cfg
@@ -151,7 +151,7 @@ cloud_providers:
     flavor_ram: ">=512"
     image:  Ubuntu-18.04
   cloudstack:
-    size: Tiny
+    size: Micro
     image: Linux Ubuntu 18.04 LTS 64-bit
     disk: 10
   vultr:
--- a/config.cfg
+++ b/config.cfg
@@ -22,6 +22,10 @@ vpn_network: 10.19.48.0/24
 vpn_network_ipv6: 'fd9d:bc11:4020::/48'
 wireguard_enabled: true
 wireguard_port: 51820
+# If you're behind NAT or a firewall and you want to receive incoming connections long after network traffic has gone silent.
+# This option will keep the "connection" open in the eyes of NAT.
+# See: https://www.wireguard.com/quickstart/#nat-and-firewall-traversal-persistence
+wireguard_PersistentKeepalive: 0
 
 # Reduce the MTU of the VPN tunnel
 # Some cloud and internet providers use a smaller MTU (Maximum Transmission
@@ -36,9 +40,9 @@ reduce_mtu: 0
 # https://wiki.strongswan.org/projects/strongswan/wiki/LoggerConfiguration
 strongswan_log_level: 2
 
-# Algo will use the following lists to block ads. You can add new block lists 
+# Algo will use the following lists to block ads. You can add new block lists
 # after deployment by modifying the line starting "BLOCKLIST_URLS=" at:
-# /usr/local/sbin/adblock.sh 
+# /usr/local/sbin/adblock.sh
 # If you load very large blocklists, you may also have to modify resource limits:
 # /etc/systemd/system/dnsmasq.service.d/100-CustomLimitations.conf
 adblock_lists:
--- a/docs/cloud-cloudstack.md
+++ b/docs/cloud-cloudstack.md
@@ -1,0 +1,20 @@
+### Configuration file
+
+You need to create a configuration file in INI format with your api key in `$HOME/.cloudstack.ini`
+
+```
+[cloudstack]
+endpoint = <endpoint>
+key = <your api key>
+secret = <your secret>
+timeout = 30
+```
+Example for Exoscale (European cloud provider exposing CloudStack API), visit https://portal.exoscale.com/u/<your account>/account/profile/api to gather the required information:
+
+```
+[cloudstack]
+endpoint = https://api.exoscale.com/compute
+key = <your api key>
+secret = <your secret>
+timeout = 30
+```
--- a/input.yml
+++ b/input.yml
@@ -20,7 +20,7 @@
       - { name: Google Compute Engine, alias: gce }
       - { name: Scaleway, alias: scaleway}
       - { name: OpenStack (DreamCompute optimised), alias: openstack }
-      - { name: CloudStack, alias: cloudstask }
+      - { name: CloudStack, alias: cloudstack }
       - { name: Install to existing Ubuntu 18.04 server (Advanced), alias: local }
   vars_files:
     - config.cfg
--- a/input.yml
+++ b/input.yml
@@ -20,6 +20,7 @@
       - { name: Google Compute Engine, alias: gce }
       - { name: Scaleway, alias: scaleway}
       - { name: OpenStack (DreamCompute optimised), alias: openstack }
+      - { name: CloudStack, alias: cloudstask }
       - { name: Install to existing Ubuntu 18.04 server (Advanced), alias: local }
   vars_files:
     - config.cfg
--- a/input.yml
+++ b/input.yml
@@ -20,7 +20,7 @@
       - { name: Google Compute Engine, alias: gce }
       - { name: Scaleway, alias: scaleway}
       - { name: OpenStack (DreamCompute optimised), alias: openstack }
-      - { name: CloudStack, alias: cloudstack }
+      - { name: CloudStack (Exoscale optimised), alias: cloudstack }
       - { name: Install to existing Ubuntu 18.04 server (Advanced), alias: local }
   vars_files:
     - config.cfg
--- a/roles/cloud-cloudstack/tasks/main.yml
+++ b/roles/cloud-cloudstack/tasks/main.yml
@@ -53,6 +53,7 @@
           ssh_key: "{{ keypair_name }}"
           security_groups: "{{ cs_security_group.name }}"
           zone: "{{ algo_region }}"
+          service_offering: "{{ size }}"
         register: cs_server
 
       - set_fact:
--- a/roles/cloud-cloudstack/tasks/main.yml
+++ b/roles/cloud-cloudstack/tasks/main.yml
@@ -6,6 +6,12 @@
     - block:
       - name: Include prompts
         import_tasks: prompts.yml
+        
+      - set_fact:
+          algo_region: >-
+            {% if region is defined %}{{ region }}
+            {%- elif _algo_region.user_input is defined and _algo_region.user_input != "" %}{{ cs_zones[_algo_region.user_input | int -1 ] }}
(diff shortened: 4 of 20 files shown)


roles/cloud-cloudstack/tasks/main.yml
```diff
@@ -0,0 +1,66 @@
...
+- block:
+    - name: Build python virtual environment
+      import_tasks: venv.yml
+
+    - block:
+      - name: Include prompts
+        import_tasks: prompts.yml
+
+      - name: Security group created
+        local_action:
```
jackivanov: Do we really have to use `local_action`?

roles/cloud-cloudstack/tasks/venv.yml
```diff
@@ -0,0 +1,15 @@
...
+  file:
+    dest: "{{ cloudstack_venv }}"
+    state: absent
+  when: clean_environment
+
+- name: Install requirements
+  pip:
+    name:
+      - cs
+      - sshpubkeys
```
jackivanov: What is `sshpubkeys` for?

roles/cloud-cloudstack/tasks/prompts.yml
```diff
@@ -0,0 +1,7 @@
+---
+- block:
+  - pause:
+      prompt: |
+        What region should the server be located in?
```
jackivanov: Can we put existing regions into this prompt? As in the other roles. Not sure if there is a modules for ansible already, but CloudStack API has a method for that https://cloudstack.apache.org/api/apidocs-4.9/apis/listRegions.html

roles/cloud-cloudstack/tasks/venv.yml
0xmilkmix: it's a requirement from the `cs_sshkeypair` module

roles/cloud-cloudstack/tasks/prompts.yml
0xmilkmix: Good point! I used the `cs` binary install alongside `cs` lib in the venv since no current `cs_*` ansible module allows to perform this for the moment.

I had to update `requirements.txt` to include `jmespath` in order to call `json_query` to parse output.

roles/cloud-cloudstack/tasks/main.yml
0xmilkmix: this is a requirement from the `cs_*` modules yes. They are designed to be run from instances as I understood  it.

jackivanov: The roles already executes on the localhost, I don't think we need this

roles/cloud-cloudstack/tasks/prompts.yml
jackivanov: We are trying to avoid updating `requirements.txt`. Jinja2 filters can handle json pretty well.

0xmilkmix: `json_query` is part of jinja2 no? What would be the other method you recommend?

jackivanov: For example you may take a loot at how it's implemented [here](https://github.com/trailofbits/algo/blob/master/roles/cloud-lightsail/tasks/prompts.yml)

roles/cloud-cloudstack/tasks/main.yml
0xmilkmix: indeed, corrected. Thank you

roles/cloud-cloudstack/tasks/prompts.yml
0xmilkmix: thanks, fixed

jackivanov: OK, thanks. We also try to not use any external clis from the playbooks. 

> shell "{{ cloudstack_venv }}/bin/cs" listZones

Could you, please, eliminate this one for now, and just add a default variable with the regions (see [this](https://github.com/trailofbits/algo/blob/master/roles/cloud-azure/defaults/main.yml#L3) for example).
Ideally we need to create an external module to discover the regions (eg: [lightsail](https://github.com/trailofbits/algo/blob/master/library/lightsail_region_facts.py)). I can do it later, after we merge this PR, or you can cover it here

input.yml
```diff
@@ -20,6 +20,7 @@
       - { name: Google Compute Engine, alias: gce }
       - { name: Scaleway, alias: scaleway}
       - { name: OpenStack (DreamCompute optimised), alias: openstack }
+      - { name: CloudStack, alias: cloudstack }
```
jackivanov: Would it be necessary to tell the user that the module is Exoscale optimised? I'm not quite familiar with CloudStack, but I think configs may vary among the providers (like in OpenStack)

roles/cloud-cloudstack/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Set facts
+        set_fact:
+          image_id: "{{ cloud_providers.cloudstack.image }}"
+          size: "{{ cloud_providers.cloudstack.size }}"
+          disk: "{{ cloud_providers.cloudstack.disk }}"
+          keypair_name: "{{ cs_keypair.name }}"
+
+      - name: Server created
+        cs_instance:
```
jackivanov: Is there any way to deploy with IPv6 by default? I saw [Exoscale has it](https://community.exoscale.com/documentation/compute/ipv6/)

docs/cloud-cloudstack.md
```diff
@@ -0,0 +1,20 @@
+### Configuration file
+
+You need to create a configuration file in INI format with your api key in `$HOME/.cloudstack.ini`
```
jackivanov: I'm also thinking about how we can integrate with an already configured environment. Eg: If someone has already configured `$HOME/.cloudstack.ini` to use another provider. 
We either may specify an environment variable `CLOUDSTACK_CONFIG` or split up the config files on [regions](https://docs.ansible.com/ansible/latest/scenario_guides/guide_cloudstack.html#regions). 
I'd prefer the method with `CLOUDSTACK_CONFIG` as it more flexible.
More details might be found [here](https://docs.ansible.com/ansible/latest/scenario_guides/guide_cloudstack.html)

jackivanov: `CLOUDSTACK_CONFIG` method:
- We add a prompt asking for a path to the config file
- We specify the variable for the whole block in the role ([example](https://docs.ansible.com/ansible/latest/scenario_guides/guide_cloudstack.html#environment-variables))

roles/cloud-cloudstack/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Keypair created
+        cs_sshkeypair:
+          name: "{{ SSH_keys.comment|regex_replace('@', '_') }}"
+          public_key: "{{ lookup('file', '{{ SSH_keys.public }}') }}"
+        register: cs_keypair
+
+      - name: Set facts
+        set_fact:
+          image_id: "{{ cloud_providers.cloudstack.image }}"
+          size: "{{ cloud_providers.cloudstack.size }}"
```
jackivanov: I just found `size` is defined, but not used. I guess it should be defined in `service_offering` for the `cs_instance` module.

config.cfg
```diff
@@ -150,6 +150,10 @@ cloud_providers:
   openstack:
     flavor_ram: ">=512"
     image:  Ubuntu-18.04
+  cloudstack:
+    size: Tiny
```
jackivanov: I think `Micro` would be enough.

roles/cloud-cloudstack/tasks/main.yml
jackivanov: Let's leave it for now

roles/cloud-cloudstack/tasks/prompts.yml
0xmilkmix: done, I've also opened an [issue](https://github.com/ansible/ansible/issues/53916) and will wait feedbacks from the module's maintainer on how to proceed.

input.yml
0xmilkmix: indeed, they seem to be the ones maintaining the ansible module :)

docs/cloud-cloudstack.md
0xmilkmix: made the modifications, let me know if it's what you had in mind

0xmilkmix:
email was not set for this repo on my host, pushing previous commits w/ local username. Fixed this from the latest commit, hope it will work for clabot

jackivanov:
@0xmilkmix It seems you need to change the author of your previous commits

✅ 0xmilkmix
❌ milkmix

roles/cloud-cloudstack/tasks/main.yml
```diff
@@ -0,0 +1,71 @@
...
+
+    - block:
+      - name: Include prompts
+        import_tasks: prompts.yml
+        
+      - set_fact:
+          algo_region: >-
+            {% if region is defined %}{{ region }}
+            {%- elif _algo_region.user_input is defined and _algo_region.user_input != "" %}{{ cs_zones[_algo_region.user_input | int -1 ]['name'] }}
+            {%- else %}{{ cs_zones[default_region | int - 1]['name'] }}{% endif %}
```
jackivanov: I tried to deploy with the default values and got this error:
> The error was: 'default_region' is undefined

roles/cloud-cloudstack/tasks/prompts.yml
```diff
@@ -0,0 +1,53 @@
...
+      prompt: |
+        What zone should the server be located in?
+          {% for z in cs_zones %}
+          {{ loop.index }}. {{ z['name'] }}
+          {% endfor %}
+
+          Enter the number of your desired zone
+          [{{ default_zone }}]
+    register: _algo_region
+  when: region is undefined
```
jackivanov: `when: region is undefined` doesn't seem to be in place. If you define `region` as an extra, var the whole block would be skipped, without asking the user where the config file is

docs/cloud-cloudstack.md
jackivanov: It seems the default value is not specified for `_cs_region`

roles/cloud-cloudstack/tasks/main.yml
0xmilkmix: fixed, typo. sorry

docs/cloud-cloudstack.md
0xmilkmix: > @0xmilkmix It seems you need to amend the author of your previous commits
> 
> ✅ 0xmilkmix
> ❌ milkmix

should be ok now

0xmilkmix:
ok... rewriting the author_email from previous commits seemed to have broken something and we now have more commits than I initially made being listed in this MR.
any git-fu ideas for this?

falzm:
I'd very much like to see this merged someday... Need any help? (disclaimer: I work @ Exoscale 😉)

0xmilkmix:
> I'd very much like to see this merged someday... Need any help? (disclaimer: I work @ Exoscale 😉)

I had to create a new branch/[pr](https://github.com/trailofbits/algo/pull/1420) to solve issues I introduced while rewriting committer email...

PR #1348, opened 2019-03-10
skip generation of SSH keypair when deploying locally

akerl:
## Description

Right now, it generates an SSH keypair in all cases. This changes that to skip if we're building locally on the system, since a keypair isn't required.

## How Has This Been Tested?

Did a local build w/ this patch, confirmed it still works w/o the keypair

## Types of changes

## Checklist:

### DIFF
--- a/playbooks/cloud-pre.yml
+++ b/playbooks/cloud-pre.yml
@@ -31,9 +31,11 @@
     size: 2048
     mode: "0600"
     type: RSA
+  when: algo_provider != "local"
 
 - name: Generate the SSH public key
   openssl_publickey:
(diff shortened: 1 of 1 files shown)


Frizz925:
Wasn't the SSH keypairs required for SSH tunnelling for each account?

akerl:
That should continue to be handled by https://github.com/trailofbits/algo/blob/master/roles/ssh_tunneling/tasks/main.yml#L61-L90

This section generates a local SSH keys for use in connecting to the cloud Algo server to provision it.

jackivanov:
Thanks!

PR #1355, opened 2019-03-12
A pause for Vultr

rodeodomino:
## Description
I've been having issues in Algo occasionally where the script will get stuck when trying to install Wireguard on Vultr. Through some experimentation, the problem was solved by forcing a pause before the Wireguard step.

## Motivation and Context
This change fixes Algo hanging on Wireguard installation when using Vultr. Maybe there's too much CPU bursting and it can't handle it? Not sure the cause, but this fixes the issue.

## How Has This Been Tested?
Ran the code on own server and ran the change through Travis CI

## Types of changes

## Checklist:


jackivanov:
I'm not a fan of hacks like this. I'd rather to debug it more detailed first. Closing the PR for now and moving to #1358

PR #1396, opened 2019-04-16
Get public IP from default interface in `install.sh`

wtgtybhertgeghgtwtg:
Currently, `publicIpFromInterface` in `install.sh` will only get the IP from the `eth0` interface, which may not exist.  Since there's a `DEFAULT_INTERFACE` defined directly above it and that isn't used anywhere else, I guess that's supposed to be used.

## Description

Get the IP from the default interface instead of always `eth0`.

## Motivation and Context

`eth0` may not be the default interface.

## How Has This Been Tested?

I ran the altered script in an environment without an `eth0` and that got set up.  I didn't see a way to run tests in `CONTRIBUTING.md`, so I guess it's supposed to go through CI.

## Types of changes

## Checklist:


jackivanov:
Thanks!

PR #1417, opened 2019-04-29
Added Restoring Option in Docs

AnderCrafter:
## Types of changes

## Checklist:


davidemyers:
I disagree with this change for several reasons:

1. Those pastebins were intended to be temporary and will expire in a year. They were to help you out of a jam but were not intended for general use.

2. If there are changes made to Algo's rules the pastebins will be out-of-date and someone will have to create new ones and update the firewall document after every change.

3. The firewall rules are generated dynamically and will be different if the user changes certain defaults in `config.cfg`.
 
4. The mention of RHEL and CentOS implies that Algo should work on those systems, but they are not supported.

PR #1429, opened 2019-05-08
Randomly generated IP address for the local dns resolver

### LINKED ISSUE #1422: DNS leak in Linux/Wireguard when LAN gateway/DNS is 172.16.0.1
**Describe the bug**

**(Update 5/6/2019)** Back to point A. Addressing conflicts between physical network and VPS-hosted DNS server indeed create DNS leaks. Even after removing this addressing conflict, DNS continues to leak secondary lookups over the physical interface upon receipt of NXDOMAIN responses. I'm observing these lookups after NXDOMAIN responses to blacklisted domains.

5/5/2019 update is moot as I was misconfiguring based on Ubuntu Server instructions that don't apply to Ubuntu Desktop because of NetworkManager conflicts.

~~**(Update 5/5/2019)** I examined this further and found that the route is not being created to `172.16.0.0/24` regardless of the LAN addressing scheme. This results in DNS leaks on all Ubuntu/Wireguard clients since the host cannot reach the VPN's DNS on `172.16.0.1`.~~

When using an Algo-provisioned VPN with the Wireguard client in Ubuntu 19.04, if the client's home network is `172.16.0.0/24` and the gateway and DNS are both `172.16.0.1` (weird edge case but it just happened to me), this will result in a DNS leak. The DNS requests will continue to flow through the LAN `172.16.0.1` because the VPN route was not established successfully for that subnet (it already existed).

The issue is a collision between Algo's hardcoded addressing scheme and the addressing scheme of my home network.

**To Reproduce**

1. Roll out a VPN service to any provider (I used Azure), default settings plus DNS ad blocker.
2. Configure an Ubuntu 19.04 client with an IP address of 172.16.0.101/24 and a gateway/DNS of 172.16.0.1.
3. Install and configure Wireguard as specified in algo/docs/client-linux-wireguard.md.
4. Start wireguard service via systemd.
5. Check https://www.whoer.net to view which DNS is detected.

**Expected behavior**

Using https://www.whoer.net to verify external IP and DNS, I expected to see N/A for my DNS. Instead I saw my ISP's DNS, indicating a leak.

**Additional context**

No additional context.

**Full log** (note changes to local addressing 5/5/2019)

```bash
user@user:~$ netstat -nr
Kernel IP routing table
Destination     Gateway         Genmask         Flags   MSS Window  irtt Iface
0.0.0.0         172.16.7.1      0.0.0.0         UG        0 0          0 enp0s31f6
169.254.0.0     0.0.0.0         255.255.0.0     U         0 0          0 enp0s31f6
172.16.7.0      0.0.0.0         255.255.255.0   U         0 0          0 enp0s31f6

user@user:~$ sudo systemctl start wg-quick@wg0

user@user:~$ netstat -nr
Kernel IP routing table
Destination     Gateway         Genmask         Flags   MSS Window  irtt Iface
0.0.0.0         172.16.7.1      0.0.0.0         UG        0 0          0 enp0s31f6
10.19.49.0      0.0.0.0         255.255.255.0   U         0 0          0 wg0
169.254.0.0     0.0.0.0         255.255.0.0     U         0 0          0 enp0s31f6
172.16.7.0      0.0.0.0         255.255.255.0   U         0 0          0 enp0s31f6
```

jackivanov:
## Description
We need to generate a random service IP for each deployment in order to avoid DNS leaking (see #1422)

## Motivation and Context
- fixes #1422
- eliminates complexity of assigning IPs for WireGuard

## How Has This Been Tested?
Deployed to EC2 and digitalocean

## Types of changes

## Checklist:

### DIFF
--- a/roles/common/templates/rules.v6.j2
+++ b/roles/common/templates/rules.v6.j2
@@ -83,7 +83,7 @@ COMMIT
 # particular virtual (tun,tap,...) or physical (ethernet) interface.
 
 # Accept DNS traffic to the local DNS resolver
--A INPUT -d fcaa::1 -p udp --dport 53 -j ACCEPT
+-A INPUT -d {{ local_service_ipv6 }}/128 -p udp --dport 53 -j ACCEPT
 
 # Drop traffic between VPN clients
 -A FORWARD -s {{ subnets|join(',') }} -d {{ subnets|join(',') }} -j {{ "DROP" if BetweenClients_DROP else "ACCEPT" }}
--- a/config.cfg
+++ b/config.cfg
@@ -100,8 +100,17 @@ dns_servers:
     - 2606:4700:4700::1111
     - 2606:4700:4700::1001
 
-# IP address for the local dns resolver
-local_service_ip: 172.16.0.1
+# Randomly generated IP address for the local dns resolver
+local_service_ip: >-
+  10.
+  {{- 255 | random(seed=algo_server_name + ansible_fqdn + 'second') }}.
+  {{- 255 | random(seed=algo_server_name + ansible_fqdn + 'third') }}.
+  {{- 255 | random(seed=algo_server_name + ansible_fqdn + 'fourth') }}
+local_service_ipv6: >-
+  FD{{ 99 | random(seed=algo_server_name + ansible_fqdn + 'first') }}:
+  {{- 9999 | random(seed=algo_server_name + ansible_fqdn + 'second') }}:
+  {{- 9999 | random(seed=algo_server_name + ansible_fqdn + 'third') }}:
+  {{- 9999 | random(seed=algo_server_name + ansible_fqdn + 'fourth') }}::1
 
 # Your Algo server will automatically install security updates. Some updates
 # require a reboot to take effect but your Algo server will not reboot itself
--- a/config.cfg
+++ b/config.cfg
@@ -40,20 +40,8 @@ wireguard_port: 51820
 wireguard_PersistentKeepalive: 0
 
 # WireGuard network configuration
-_wireguard_network_ipv4:
-  subnet: 10.19.49.0
-  prefix: 24
-  gateway: 10.19.49.1
-  clients_range: 10.19.49
-  clients_start: 2
-_wireguard_network_ipv6:
-  subnet: 'fd9d:bc11:4021::'
-  prefix: 48
-  gateway: 'fd9d:bc11:4021::1'
-  clients_range: 'fd9d:bc11:4021::'
-  clients_start: 2
-wireguard_network_ipv4: "{{ _wireguard_network_ipv4['subnet'] }}/{{ _wireguard_network_ipv4['prefix'] }}"
-wireguard_network_ipv6: "{{ _wireguard_network_ipv6['subnet'] }}/{{ _wireguard_network_ipv6['prefix'] }}"
+wireguard_network_ipv4: 10.19.49.0/24
+wireguard_network_ipv6: fd9d:bc11:4021::/48
 
 # Reduce the MTU of the VPN tunnel
 # Some cloud and internet providers use a smaller MTU (Maximum Transmission
@@ -101,16 +89,8 @@ dns_servers:
     - 2606:4700:4700::1001
 
 # Randomly generated IP address for the local dns resolver
-local_service_ip: >-
-  10.
-  {{- 255 | random(start=20, seed=algo_server_name + ansible_fqdn + 'second') }}.
-  {{- 255 | random(seed=algo_server_name + ansible_fqdn + 'third') }}.
-  {{- 255 | random(seed=algo_server_name + ansible_fqdn + 'fourth') }}
-local_service_ipv6: >-
-  FD{{ 99 | random(seed=algo_server_name + ansible_fqdn + 'first') }}:
-  {{- 9999 | random(seed=algo_server_name + ansible_fqdn + 'second') }}:
-  {{- 9999 | random(seed=algo_server_name + ansible_fqdn + 'third') }}:
-  {{- 9999 | random(seed=algo_server_name + ansible_fqdn + 'fourth') }}::1
+local_service_ip: "{{ '172.16.0.1' | ipmath(1048573 | random(seed=algo_server_name + ansible_fqdn)) }}"
+local_service_ipv6: "{{ 'fd00::1' | ipmath(1048573 | random(seed=algo_server_name + ansible_fqdn)) }}"
 
 # Your Algo server will automatically install security updates. Some updates
 # require a reboot to take effect but your Algo server will not reboot itself
--- a/config.cfg
+++ b/config.cfg
@@ -103,7 +103,7 @@ dns_servers:
 # Randomly generated IP address for the local dns resolver
 local_service_ip: >-
   10.
-  {{- 255 | random(seed=algo_server_name + ansible_fqdn + 'second') }}.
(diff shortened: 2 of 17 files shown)


roles/common/templates/rules.v6.j2
```diff
@@ -83,7 +83,7 @@ COMMIT
 # particular virtual (tun,tap,...) or physical (ethernet) interface.
 
 # Accept DNS traffic to the local DNS resolver
--A INPUT -d fcaa::1 -p udp --dport 53 -j ACCEPT
+-A INPUT -d {{ local_service_ipv6 }}/128 -p udp --dport 53 -j ACCEPT
```
rodeodomino:
```suggestion
-A INPUT -d {{ local_service_ipv6 }} -p udp --dport 53 -j ACCEPT
```

TC1977:
Probably need to change the IP address ranges in https://github.com/trailofbits/algo/blob/master/roles/cloud-ec2/defaults/main.yml, https://github.com/trailofbits/algo/blob/master/roles/cloud-ec2/files/stack.yaml, and documented https://github.com/trailofbits/algo/blob/master/docs/client-linux-wireguard.md, as well as the documented DNS workaround [in the faq](https://github.com/trailofbits/algo/blob/master/docs/faq.md#can-dns-filtering-be-disabled).

Also, won't changing the DNS resolver to a random 10.x.x.x IP address introduce the 1/32512 chance that it lands in the 10.19.49.x or 10.19.48.x range and screws with Wireguard or IPsec?

roles/common/templates/rules.v6.j2
jackivanov: No, because `local_service_ipv6` is the destination address that we should allow to talk to, not the source

jackivanov:
@TC1977 EC2 cidr blocks are different from the service IP and used only inside the VPC, so we don't need to change variables, but indeed, we need to update the docs and the faq, I'll do it shortly, thanks!

roles/common/templates/rules.v6.j2
shapiro125: interesting! cool, thank you.

rodeodomino: interesting! cool thank you

TC1977:
Successfully cloud installed onto Vultr from macOS 10.14.4, with new random DNS resolver IP, and confirmed the Wireguard iOS client works:

```
ok: [173.199.123.37] => {
    "msg": [
        [
            "\"#                          Congratulations!                            #\"", 
            "\"#                     Your Algo server is running.                     #\"", 
            "\"#    Config files and certificates are in the ./configs/ directory.    #\"", 
            "\"#              Go to https://whoer.net/ after connecting               #\"", 
            "\"#        and ensure that all your traffic passes through the VPN.      #\"", 
            "\"#                     Local DNS resolver 10.229.197.10                    #\"", 
            ""
        ], 
        "    \"#        The p12 and SSH keys password for new users is NyqxmuFb       #\"\n", 
        "    \"#        The CA key password is 4c0d6b38adf6c411e4ddb72588a6d23e       #\"\n", 
        "    \"#      Shell access: ssh -i configs/algo.pem root@173.199.123.37        #\"\n"
    ]
}

```

Local installation onto a Vultr Ubuntu 18.04 instance also successful, and Wireguard iOS client works:

```
ok: [localhost] => {
    "msg": [
        [
            "\"#                          Congratulations!                            #\"", 
            "\"#                     Your Algo server is running.                     #\"", 
            "\"#    Config files and certificates are in the ./configs/ directory.    #\"", 
            "\"#              Go to https://whoer.net/ after connecting               #\"", 
            "\"#        and ensure that all your traffic passes through the VPN.      #\"", 
            "\"#                     Local DNS resolver 10.61.189.116                    #\"", 
            ""
        ], 
        "    \"#        The p12 and SSH keys password for new users is 0NoB9z77       #\"\n", 
        "    \"#        The CA key password is 5971ff0378de844a11063c08cc169e4e       #\"\n", 
        "    "
    ]
}
```

davidemyers:
@jackivanov I've rewritten the Linux WireGuard client instructions as shown below to abandon the approach of trying to use `systemd-resolved` for DNS and instead use `openresolv` to rewrite `/etc/resolv.conf`. If I submit a PR now it will end up in conflict with this one. How would you prefer to proceed:

1. Should I wait for this PR to be merged before submitting another one?
2. Do you want to remove your changes to `client-linux-wireguard.md` from this PR and I'll submit a PR that won't conflict?
3. Do you want me to somehow get you the source to the text below for inclusion in this PR?

Whatever works best for you. Thanks.

----------

# Using Ubuntu as a Client with WireGuard

## Install WireGuard

To connect to your AlgoVPN using [WireGuard](https://www.wireguard.com) from Ubuntu, first install WireGuard:

```shell
# Add the WireGuard repository:
sudo add-apt-repository ppa:wireguard/wireguard

# Update the list of available packages (not necessary on 18.04 or later):
sudo apt update

# Install the tools and kernel module:
sudo apt install wireguard openresolv
```

For installation on other Linux distributions, see the [Installation](https://www.wireguard.com/install/) page on the WireGuard site.

## Locate the Config File

The Algo-generated config files for WireGuard are named `configs/<ip_address>/wireguard/<username>.conf` on the system where you ran `./algo`. One file was generated for each of the users you added to `config.cfg`. Each WireGuard client you connect to your AlgoVPN must use a different config file. Choose one of these files and copy it to your Linux client.

## Configure WireGuard

Finally, install the config file on your client as `/etc/wireguard/wg0.conf` and start WireGuard:

```shell
# Install the config file to the WireGuard configuration directory on your
# Linux client:
sudo install -o root -g root -m 600 <username>.conf /etc/wireguard/wg0.conf

# Start the WireGuard VPN:
sudo systemctl start wg-quick@wg0

# Check that it started properly:
sudo systemctl status wg-quick@wg0

# Verify the connection to the AlgoVPN:
sudo wg

# See that your client is using the IP address of your AlgoVPN:
curl ipv4.icanhazip.com

# Optionally configure the connection to come up at boot time:
sudo systemctl enable wg-quick@wg0
```

If your Linux distribution does not use `systemd` you can bring up WireGuard with `sudo wg-quick up wg0`.

jackivanov:
@davidemyers I've included the docs, thanks!

jackivanov:
I revised the whole method of allocating IPs, and made some changes. We need to do more tests

davidemyers:
IPv6 appears to be broken in WireGuard. The server-side IPv6 `AllowedIPs` are wrong:

```
[Interface]
Address = 10.19.49.1/24 ,fd9d:bc11:4021::1/48
ListenPort = 51820
PrivateKey = 4KVwPJFtlvly172KG+fPjVMg7oW3Rkun85+ffs9fL1g=
SaveConfig = false


[Peer]
# phone
PublicKey = nnU54wNnFrm9B/JuZm7U7owLNcYoTHn/OxcsgOhLP0g=
AllowedIPs = 10.19.49.2/32,fd9d:bc11:4021::1/128

[Peer]
# laptop
PublicKey = 6aDQd5kJjqkqP2cALCRyUJqPLkHsABMGLrDHKCYFGEY=
AllowedIPs = 10.19.49.3/32,fd9d:bc11:4021::2/128

[Peer]
# desktop
PublicKey = bK2drxn7oah4x/OGfhA+L6r0gtGVgUxWatV8CRFQ3EM=
AllowedIPs = 10.19.49.4/32,fd9d:bc11:4021::3/128
```

davidemyers:
This now works for me on DigitalOcean using WireGuard on iOS and Ubuntu Disco Desktop.

PR #1441, opened 2019-05-20
Update users: add server pick-list

### LINKED ISSUE #1438: Have update-users create a list of servers
Running the update-users script could be easier. I ran it on a local install and simply hit return, rather than entering `localhost`, which caused a failure. People who have multiple servers to maintain could have issues remembering IP addresses or domain names.

Specifically, the first prompt `Enter the IP address of your server: (or use localhost for local installation)` could instead give a menu of servers that have been installed, including localhost if there's been a local install. I think the easiest way to be to have it parse the results of `shell: ls configs/`, and either drop `algo.pem` and `algo.pem.pub`, or look only for directories. But I have no idea how to do that.

Screen Shot 2019-05-17 at 12 01 51 PM

TC1977:
## Description

When updating users, you no longer have to type out the entire IP address or domain name (or 'localhost') of the Algo server that you deployed and want to update. This might also be helpful if you've deployed multiple Algo servers.

## Motivation and Context

Closes #1438. Typing a whole IP address is hard. Cutting-and-pasting is also hard. Typing a number is easy.

## How Has This Been Tested?

Numerous runs of `ansible-playbook users.yml`.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/users.yml
+++ b/users.yml
@@ -16,16 +16,24 @@
             patterns: ".config.yml"
           register: _configs_list
 
+        - name: Verify servers
+          assert:
+            that: _configs_list.matched > 0
+            msg: No servers found, nothing to update.
+
         - name: Build string of installed servers
           set_fact:
-            server_list: "{% if server_list is defined %}{{ server_list }},{% endif %}{{ item.path|replace('configs/','')|replace('/.config.yml','') }}"
-          with_items: "{{ _configs_list.files }}"
+            server_list: >-
+              [{% for i in _configs_list.files %}
+                '{{ i.path.split('/')[1] }}'
+                {{ ',' if not loop.last else '' }}
+              {% endfor %}]
 
         - name: Server address prompt
           pause:
             prompt: |
               Select the server to update user list below:
-                {% for r in server_list.split(',') %}
+                {% for r in server_list %}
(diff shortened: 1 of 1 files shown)


users.yml
```diff
@@ -7,17 +7,30 @@
 
   tasks:
     - block:
+        - name: Get list of installed servers
```
jackivanov: Let's move to the native module. Something like:
```yaml
- name: Get list of installed servers
  find:
    paths: configs/
    depth: 2
    recurse: true
    hidden: true
    patterns: ".config.yml"
  register: configs
```

TC1977: I'd thought the `find` module only accepted absolute paths.

jackivanov:
Should we add additional info to the list, like cloud, region and server name? Something like:

```
Select the server to update user list below:
  1. DigitalOcean   fra1        test-server-01    206.189.19.82
  2. EC2            eu-west-1   prod-server-01    18.203.231.98
```

TC1977:
I don't need it myself, but I think it would be really useful for people running multiple servers.

jackivanov:
let's leave it as is for now

PR #1465, opened 2019-06-02
Adding preshared key support

elreydetoda:
## Description
So I in this pull request I simply took the existing ansible code and extended it to include preshared key support for wireguard. This is what it says about preshared keys in the man page:
```
PresharedKey — a base64 preshared key generated by wg genpsk. Optional, and may be omitted. This option adds an additional layer of symmetric-key cryptography to be mixed into the already existing public-key cryptography, for post-quantum resistance.
```
so since it is as simple as adding a `wg genpsk` instead of `wg genkey` and adding it to the configs, and you are supposed to get an extra layer of encryption for pretty much nothing.

## Motivation and Context
the project touts that it
> It uses the most secure defaults available

and I never see any other tutorials mention about this feature at all. So I figured maybe you didn't know about it and though why not include a little more security for a minimal amount of effort. I just wanted to help out since I saw it didn't include it yet :smile: 

## How Has This Been Tested?

I ran it multiple times from a vagrant box image to a digital ocean vps: https://gist.github.com/elreydetoda/bac472aef59fc8d47145c4f25330187b#file-vagrantfile-elrey

simply ran `vagrant up` and then `./algo`

## Types of changes

## Checklist:

it is literally copied and pasted of what was already there.

I don't believe it needs documentation because it does this transparently to the user.

### DIFF
--- a/roles/wireguard/tasks/keys.yml
+++ b/roles/wireguard/tasks/keys.yml
@@ -1,16 +1,7 @@
 ---
-- name: Delete the private lock files
+- name: Delete the lock files
   file:
     dest: "{{ config_prefix|default('/') }}etc/wireguard/private_{{ item }}.lock"
-    state: absent
-  when: keys_clean_all|bool
-  with_items:
-    - "{{ users }}"
-    - "{{ IP_subject_alt_name }}"
-    
-- name: Delete the preshared lock files
-  file:
-    dest: "{{ config_prefix|default('/') }}etc/wireguard/preshared_{{ item }}.lock"
     state: absent
   when: keys_clean_all|bool
   with_items:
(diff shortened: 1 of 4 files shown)


elreydetoda:
The travis-ci errors are things that I haven't touched. Just a suggestion for merging (not saying it will, but if it does) I would suggest to squash merge and just keep the same title.

also, thank you so much for making this project. I was going to stand up my own wireguard server but saw how this project allows you to add the dnssec, ad blocking, etc... and I hadn't even thought of doing that! So, thank you so much!
I saw about this project from here (just for reference): https://clo.ng/blog/algo_vpn/

elreydetoda:
The travis-ci errors are things that I haven't touched. Just a suggestion for merging (not saying it will, but if it does) I would suggest to squash merge and just keep the same title.

also, thank you so much for making this project. I was going to stand up my own wireguard server but saw how this project allows you to add the dnssec, ad blocking, etc... and I hadn't even thought of doing that! So, thank you so much!
I saw about this project from here (just for reference): https://clo.ng/blog/algo_vpn/

davidemyers:
This worked for me when deploying to DigitalOcean.

I've also wondered if using the WireGuard PSK would be beneficial. I'm interested to know what others think.

TC1977:
As per that page, “WireGuard also supports an optional pre-shared key that is mixed into the public key cryptography. When pre-shared key mode is not in use, the pre-shared key value used below is assumed to be an all-zero string of 32-bytes.”

So Algo still uses the most secure _default_ option available. 😅

On a more serious note, does this really increase security? The client PSKs are still available on the client machine and deploy machine, and the server PSKs are still available on the server, just like the other keys. I think there’s an easier way to attack the server than renting out a quantum computer.

elreydetoda:
Haha, nice find. I had missed that when skimming before :laugh: Ya, @TC1977 I am sure there are other cheaper ways, but it doesn't hurt (that I know of) since it is already built in.

jackivanov:
@elreydetoda I'm not allowed to push to your branch, so, could you, please, rebase form the master yourself to fix the tests?

elreydetoda:
> @elreydetoda I'm not allowed to push to your branch, so, could you, please, rebase form the master yourself to fix the tests?

Thanks for the fix, and looks like we are good :smile:

elreydetoda:
Is there anything else that needs to happen with this @jackivanov?

PR #1477, opened 2019-06-09
Changing adblock from local to server

rodeodomino:
## Description
Changing the adblock script to create a list of `server=` instead of `local=`.

## Motivation and Context
Honestly not sure if this is necessary, but I was looking into whether this adblock method worked with wildcards (example.com and a.example.com both blocked) and most tutorials mention using `server=` for ad blocking.

Reference: https://bering-uclibc.zetam.org/wiki/Bering-uClibc_6.x_-_User_Guide_-_Advanced_Topics_-_Setting_Up_Ad_blocking_with_dnsmasq

## How Has This Been Tested?
Tested on Amazon EC2 and works as expected (but not sure which is the better/safer/more secure way of doing it.

## Types of changes

## Checklist:

- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/roles/dns_adblocking/templates/adblock.sh.j2
+++ b/roles/dns_adblocking/templates/adblock.sh.j2
@@ -25,7 +25,7 @@ then
 fi
 
 #Sort the download/black lists
-awk '/^[^#]/ { print "local=/" $1 "/" }' "$TEMP" | sort -u > "$TEMP_SORTED"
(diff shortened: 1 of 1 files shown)


jackivanov:
I don't think it's necessary, they both work like synonyms.

rodeodomino:
Thanks for the clarifications!

PR #1490, opened 2019-06-20
Add default IPv6 routes to the windows powershell script

### LINKED ISSUE #1453: Windows IKEv2 over IPSec doesn't provide a IPv6 connection
**Describe the bug**

When connecting via IPSec to the VPN on Windows where my physical network is IPv4 only, I do not end up with a dual-stack connection.

**To Reproduce**

1. ./algo
2. Install .ps1 file for Windows VPN
3. Connect using Settings

**Expected behavior**

A dual-stack connection after routing through the VPN.

**Additional context**

As above, my physical network is IPv4-only. I DO get an IPv6 address using Wireguard on my Android device (kernel-level [rooted] wireguard) and my iPad (iOS wireguard client).

image

jackivanov:
## Description
Windows doesn't add IPv6 routes for IKEv2 connections automatically, so we should add them via the powershell script. For some reason ::/0 doesn't work and we have to use ::/1 + 8000::/1

## Motivation and Context
Fixes #1453

## How Has This Been Tested?
Deployed an instance with windows support and configured my windows vm successfully 

## Types of changes

## Checklist:

### DIFF
--- a/roles/strongswan/templates/client_windows.ps1.j2
+++ b/roles/strongswan/templates/client_windows.ps1.j2
@@ -174,6 +174,12 @@ function Add-AlgoVPN {
     }
     Add-VpnConnection @addVpnParams
 
+    $addVpnRouteParams = @{
+        ConnectionName = $VpnName
+    }
+    Add-VpnConnectionRoute @addVpnRouteParams -DestinationPrefix ::/1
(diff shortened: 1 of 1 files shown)


jackivanov:
We've decided to deprecate IKEv2 for windows https://github.com/trailofbits/algo/issues/1493

PR #1496, opened 2019-06-21
PKI to tmpfs

### LINKED ISSUE #145: tmpfs for key generation
In my mind Algo is not really designed to persist. When you spin it up when you need it and tear it down when you are done. When you need it again you spin up a new instance, with new keys. 

If everyone agrees that statement aligns with the goals of Algo I propose we use tmpfs as a memory resident only storage of sensitive keys as a layer of protection against postmortem forensics analysis that could be used for key recovery if the image were ever recovered. While this may be an unlikely event, the level of effort to implement is minimal.

jackivanov:
## Description
As discussed in #145. The config files will be stored in tmpfs only if the `store_pki` (retain the PKI) is not chosen, and unmounted once the server is ready. MacOS implementation might be a bit dirty, but I didn't find any other way, so we need to review whether this would be appropriate for merging.

## Motivation and Context
Closes #145

## How Has This Been Tested?
Deployed a few times without errors

## Types of changes

## Checklist:


playbooks/tmpfs/macos.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: MacOS | set OS specific facts
+  set_fact:
+    tmpfs_volume_name: "AlgoVPN-{{ IP_subject_alt_name }}"
+    tmpfs_volume_path: /Volumes
+
+- name: MacOS | mount a ram disk
+  shell:
+    diskutil info "/{{ tmpfs_volume_path }}/{{ tmpfs_volume_name }}/" ||
+    diskutil erasevolume HFS+ "{{ tmpfs_volume_name }}" $(hdiutil attach -nomount ram://64000)
```
disconnect3d: Shouldn't we put the full volume path here, as in we do to `diskutil info`?

playbooks/cloud-post.yml
```diff
@@ -37,6 +37,14 @@
     state: present
   when: cloud_instance_ip != "localhost"
 
+- name: Mount tmpfs
+  import_tasks: tmpfs/main.yml
+  when:
+    - pki_in_tmpfs
+    - not algo_store_pki
+    - ansible_system == "Darwin" or
+      ansible_system == "Linux"
```
disconnect3d: Can algo be used on machines that doesn't have much memory? Do people do that?

I mean, it might be that either `/dev/shm` won't exist on Linux (https://superuser.com/questions/45342/when-should-i-use-dev-shm-and-when-should-i-use-tmp) or that we won't have enough memory to use it.

Maybe we should check for those two conditions? Not sure how to determine the required memory, maybe we could skip that or just have a minimum required.

playbooks/tmpfs/macos.yml
disconnect3d: Also: can it be that diskutil won't be in PATH (https://superuser.com/questions/213088/diskutil-command-not-found-in-os-x-terminal)? Maybe we should handle this?

playbooks/tmpfs/umount.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Linux | Delete the PKI directory
+  file:
+    path: "/{{ facts.tmpfs_volume_path }}/{{ facts.tmpfs_volume_name }}/"
+    state: absent
+  when: facts.ansible_system == "Linux"
+
+- name: MacOS | eject the ram disk
+  command: diskutil eject "/{{ facts.tmpfs_volume_path }}/{{ facts.tmpfs_volume_name }}/" || true
```
disconnect3d: What can be the reasons the eject fails? It seems that the `|| true` will silently ignore if it does?

playbooks/tmpfs/macos.yml
jackivanov: > Shouldn't we put the full volume path here, as we do when diskutil info is used?

No, `eraseVolume` requires volume name

> Also: can it be that diskutil won't be in PATH (https://superuser.com/questions/213088/diskutil-command-not-found-in-os-x-terminal)? Maybe we should handle this?

Right. I expanded the command to use the full path instead

playbooks/tmpfs/umount.yml
jackivanov: Yes, it does. I've separated the task and removed `|| true`. With this approach no errors expected.

playbooks/cloud-post.yml
jackivanov: I'll add a check to ensure that `/dev/shm` is available. Algo doesn't require much memory to store files, and rather Ansible fails if there is not enough memory available.

jackivanov: Actually Ansible doesn't even run without `/dev/shm` being mounted. So, I would rather just add  these cases to the troubleshooting guide.

PR #1498, opened 2019-06-22
Troubleshooting IPsec NAT issues

### LINKED ISSUE #520: Router NAT settings can cause client reconnection issues
Here's a description of an issue I've found written in a form suitable for the Troubleshooting document.

I have multiple devices and they have problems staying connected

Check your router’s NAT settings.  In pfSense, NAT randomizes the source ports of outgoing connections. In order to accommodate IPsec clients whose IKE implementations don’t support NAT-T, pfSense exempts outgoing IKE connections (port 500) from randomization and keeps the source port static. This creates a limitation of only one current IKE negotiation at a time. Any subsequent attempt to create an IPsec connection fails until the firewall state created by the previous attempt expires (in about a minute).

pfSense users can avoid this issue by disabling the static port NAT exemption for IKE. In Outgoing NAT settings, switch to Manual Outbound NAT and disable the rule for traffic from your LAN address to port 500.

This issue is more likely to be seen by those with multiple iOS clients as they tend to attempt frequent reconnections when sleeping.

davidemyers:
## Description

Add a section on troubleshooting IPsec NAT issues.

## Motivation and Context

Closes #520.

## How Has This Been Tested?

Documentation only.

## Types of changes

- [x] Documentation

## Checklist:


jackivanov:
Thanks!

PR #1512, opened 2019-07-08
adding f-droid.org as resource for android clients

leinadlime:
## Description
adding link to F-Droid.org in docs/android-client.md 

## Motivation and Context
Promoting free software repositories.
Lets include also a link to the the FOSS repository f-droid,
some user may like it better or not know it is there,..

## How Has This Been Tested?
no. it's just text for the documentation

## Types of changes

## Checklist:

### DIFF
--- a/docs/client-android.md
+++ b/docs/client-android.md
@@ -2,5 +2,5 @@
 
 ## Installation via profiles
 
-1. [Install the WireGuard VPN Client](https://play.google.com/store/apps/details?id=com.wireguard.android).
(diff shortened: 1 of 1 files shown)


davidemyers:
I'm not an Android user so perhaps I don't understand how things work, but I think it's important that users are directed to install the genuine WireGuard apps as built and submitted by the WireGuard Team. The version of WireGuard in the Google Play Store appears to be from the WireGuard Team but the one on F-Droid does not. Doesn't that represent a potential security risk?

dguido:
Thanks, but we don't recommend using third-party app stores for the reasons David mentioned above.

PR #1549, opened 2019-08-15
Refactor to support Ansible 2.8

### LINKED ISSUE #988: Add Hetzner Cloud
Please implement support for [Hetzner Cloud](https://www.hetzner.com/cloud?country=ot). They have very competitive pricing and a lot of bandwidth included.

[They have an API](https://docs.hetzner.cloud/)

This is their smallest server instance:
image

### LINKED ISSUE #1502: Instance creation issue(Scaleway)
The error occurs after specifying the api key (log below). The problem occurs only if i'm changing the parameter "size" in the file "config.cfg":
DEFAULT

```
  scaleway:
    size: START1-S
    image: Ubuntu Bionic Beaver
    arch: x86_64
```

EDITED

```
  scaleway:
    size: DEV1-S
    image: Ubuntu Bionic Beaver
    arch: x86_64
```
DEV1-S is available and can be created manually

**LOG**

```
TASK [cloud-scaleway : Set scaleway facts] **************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [cloud-scaleway : Gather Scaleway organizations facts] *********************************************************************************************************************************************************************************
ok: [localhost]

TASK [cloud-scaleway : Get images] **********************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [cloud-scaleway : Set cloud specific facts] ********************************************************************************************************************************************************************************************
ok: [localhost]

TASK [cloud-scaleway : Create a server] *****************************************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'algo_instance.msg.public_ip' failed. The error was: error while evaluating conditional (algo_instance.msg.public_ip): 'ansible.utils.unsafe_proxy.AnsibleUnsafeText object' has no attribute 'public_ip'"}

TASK [include_tasks] ************************************************************************************************************************************************************************************************************************
included: /home/ss/algo-master/playbooks/rescue.yml for localhost

TASK [debug] ********************************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "fail_hint": [
        "Sorry, but something went wrong!",
        "Please check the troubleshooting guide.",
        "https://trailofbits.github.io/algo/troubleshooting.html"
    ]
}
```

### LINKED ISSUE #1571: Digital Ocean Droplet creation failed with DoError: found unpermitted parameters: virtio
Perhaps DO API has changed? Error output:

TASK [cloud-digitalocean : Creating a droplet...] ******************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: DoError: found unpermitted parameters: virtio
fatal: [localhost]: FAILED! => {"changed": false, "msg": "found unpermitted parameters: virtio"}
included: /home/beoberha/algo/playbooks/rescue.yml for localhost

TASK [debug] *******************************************************************
ok: [localhost] => {
    "fail_hint": [
        "Sorry, but something went wrong!",
        "Please check the troubleshooting guide.",
        "https://trailofbits.github.io/algo/troubleshooting.html"
    ]
}

TASK [Fail the installation] ***************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed as requested from task"}

PLAY RECAP *********************************************************************
localhost                  : ok=32   changed=6    unreachable=0    failed=2

jackivanov:
## Upgrades and Fixes
- Upgrade to Ansible 2.8.3
- Upgrade to Python 3
- Big documentation changes  

## Cloud provider changes
- Add a new cloud provider - Hetzner Cloud. (Fixes #988)
- Fix Scaleway instance creating issue. (Fixes #1502)
- Remove deprecated DigitalOcean (Fixes #1571) and GCE modules

## Breaking changes
- python virtual environment location is now `.env`
- GCE: from now, the installer expects you to specify a region (eg: us-east1), instead of a zone (eg: us-east1-a).

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


davidemyers:
I tried a few tests. Starting with a new Ubuntu Server 19.04 container I ran cloud deployments to Vultr and DigitalOcean. Then I created an Ubuntu Server 19.04 droplet on DigitalOcean and did a local install. All tested fine using WireGuard on iOS.

jackivanov:
Thanks @phaer! Somehow I missed the api tokens. I just sent another solution for the apt lock, works well.

phaer:
> I just sent another solution for the apt lock, works well.
Ah yes, thanks! My ansible skills are a bit rusty, didn't think about that.

jackivanov:
Scaleway modules seems to be all broken.

```
TASK [cloud-scaleway : Gather Scaleway organizations facts] ************************************************************************************************************************************************
 [WARNING]: https://account.scaleway.com/organizations?

ok: [localhost]

TASK [cloud-scaleway : Get images] *************************************************************************************************************************************************************************
 [WARNING]: https://cp-par1.scaleway.com/images?

ok: [localhost]

TASK [cloud-scaleway : Set cloud specific facts] ***********************************************************************************************************************************************************
ok: [localhost]

TASK [cloud-scaleway : Create a server] ********************************************************************************************************************************************************************
 [WARNING]: https://cp-par1.scaleway.com/servers?name=scw-test&per_page=1?

fatal: [localhost]: FAILED! => {"changed": false, "msg": "Error during server search: (400) {u'fields': {u'per_page': [u'expected int']}, u'message': u'Validation Error', u'type': u'invalid_request_error'}"}
```

@remyleone Is this something you can take a look at?

davidemyers:
An Ubuntu Server 18.04 droplet local install works OK too.

davidemyers:
Also, just for "fun", an Ubuntu Server 19.04 droplet local install works if you substitute `python` with `python3` in the setup instructions. :smile:

TC1977:
Update docs?

jackivanov:
@TC1977 I would appreciate your help :) I guess, you can commit directly to this branch?

TC1977:
I can't commit directly, but I can submit a PR to this branch.

jackivanov:
Yes, that's OK

TC1977:
Ok, I think I'm all set. #1552

davidemyers:
A cloud deployment  from an Ubuntu Server 19.04 container to DigitalOcean using `python3` works with these new DO modules.

It works if `LANG=C.UTF-8` but fails if `LANG="en_US.UTF-8"` on this bit where the p12 password is generated:

```
openssl rand 8 |
      python -c 'import sys,string; chars=string.ascii_letters + string.digits + "_@"; print("".join([chars[ord(c) % 64] for c in list(sys.stdin.read())]))'
```

davidemyers:
A cloud deployment from macOS 10.14.6 to DigitalOcean using `python3` and `LANG=C.UTF-8` also works.

Since I already had Homebrew installed I set up the dependencies like this:
```
brew install python
python3 -m pip install --upgrade virtualenv # Don't need --user
```
I think when the time comes to move to `python3` the instructions for macOS users might want to include both using Homebrew and installing the official Python 3 package as options. Homebrew is convenient but if you're not already using it it's kind of a big dependency. Though I wonder if the official Python 3 package will require updating the certificate store like the Python 2 package does, as discussed in #1545.

jackivanov:
@davidemyers I think we need to address python3 in a separate issue

davidemyers:
@jackivanov I agree, I was just curious to know what already worked and thought I'd record what I found in case it's useful later.

jackivanov:
Actually, I'd like to switch to python3 in this pr, so still work in progress

davidemyers:
I'm not a Python expert, but I think [PEP 538](https://www.python.org/dev/peps/pep-0538/) explains the issue with generating the p12 password in Python 3. This works at the command line on 18.04, 19.04, and macOS:
```
openssl rand 8 | LANG=C python3 -c 'import sys,string; chars=string.ascii_letters + string.digits + "_@"; print("".join([chars[ord(c) % 64] for c in list(sys.stdin.read())]))'
```
I haven't tried it this way yet when invoked from Ansible.

davidemyers:
`algo` needs `env` -> `.env`. After being reverted, I guess.

jackivanov:
@davidemyers Did you use [this guide](https://github.com/trailofbits/algo/blob/a54aab66e0199c2973f0db4ab06cb1a876ae41e9/docs/deploy-from-fedora-workstation.md)?

davidemyers:
@jackivanov No, I ignored the guide, I wanted to see what the minimum requirements really were.
All I did was (as `root`):
```
dnf upgrade
reboot
dnf install python3-virtualenv
```
Then, after cloning this branch:
```
python3 -m virtualenv --python="$(which python3)" --system-site-packages .env
source .env/bin/activate
python3 -m pip install -U pip virtualenv
python3 -m pip install -r requirements.txt
```

davidemyers:
I've attempted to rewrite the installation section of the README. I haven't actually tested using the official Python package on macOS so that section might not be correct. Comments?

-----
## Deploy the Algo Server

The easiest way to get an Algo server running is to run it on your local system and let it set up a _new_ virtual machine in the cloud for you.

1. **Setup an account on a cloud hosting provider.** Algo supports [DigitalOcean](https://m.do.co/c/4d7f4ff9cfe4) (most user friendly), [Amazon Lightsail](https://aws.amazon.com/lightsail/), [Amazon EC2](https://aws.amazon.com/), [Vultr](https://www.vultr.com/), [Microsoft Azure](https://azure.microsoft.com/), [Google Compute Engine](https://cloud.google.com/compute/), [Scaleway](https://www.scaleway.com/), [DreamCompute](https://www.dreamhost.com/cloud/computing/) or other OpenStack-based cloud hosting, [Exoscale](https://www.exoscale.com) or other CloudStack-based cloud hosting,  or [Hetzner Cloud](https://www.hetzner.com/).

2. **Get a copy of Algo.** The Algo scripts will be installed on your local system. There are two ways to get a copy:

	- Download the [ZIP file](https://github.com/trailofbits/algo/archive/master.zip). Unzip the file to create a directory named `algo-master` containing the Algo scripts.
	- Run the command `git clone https://github.com/trailofbits/algo.git` to create a directory named `algo` containing the Algo scripts.

3. **Install Algo's core dependencies.** Algo requires that Python 3 and one supporting package are installed on your system.

    - **macOS:** Apple does not provide Python 3 with macOS. There are two ways to obtain it:
		* Use the [Homebrew](https://brew.sh) package manager. After installing Homebrew, run these commands from Terminal:
		```bash
			brew install python3
			python3 -m pip install --upgrade virtualenv
		```
		* Download and install the latest stable [Python 3 package](https://www.python.org/downloads/mac-osx/) and run the included *Install Certificates* command from Finder. Then from Terminal run:
		```bash
			python3 -m pip install --user --upgrade virtualenv
		```

    - **Linux:** Recent releases of Ubuntu, Debian, and Fedora come with Python 3 already installed. Make sure your system is up-to-date and install the needed supporting package:
    	* Ubuntu and Debian:
    	```bash
    		sudo apt update && sudo apt upgrade -y
    		sudo apt install -y python3-virtualenv
    	```
    	* Fedora:
    	```bash
    		sudo dnf upgrade -y
    		sudo dnf install -y python3-virtualenv
    	```
    	* Red Hat and CentOS: See this [documentation](docs/deploy-from-redhat-centos6.md).

     - **Windows:** Use the Windows Subsystem for Linux (WSL) to create your own copy of Ubuntu running under Windows from which to install and run Algo. See the [Windows documentation](docs/deploy-from-windows.md).

4. **Install Algo's remaining dependencies.** You'll need to run these commands from the Algo directory each time you download a new copy of Algo. In a Terminal window `cd` into the `algo-master` (ZIP file) or `algo` (`git clone`) directory and run:
    ```bash
    python3 -m virtualenv --python="$(command -v python3)" .env &&
        source .env/bin/activate &&
        python3 -m pip install -U pip virtualenv &&
        python3 -m pip install -r requirements.txt
    ```
     On Fedora add the option `--system-site-packages` to the first command above. On macOS install the C compiler if prompted.

5. **List the users to create.** Open the file `config.cfg` in your favorite text editor. Specify the users you wish to create in the `users` list. Create a unique user for each device you plan to connect to your VPN. If you want to be able to add or delete users later, you **must** select `yes` at the `Do you want to retain the keys (PKI)?` prompt during the deployment.

6. **Start the deployment.** Return to your terminal. In the Algo directory, run `./algo` and follow the instructions. There are several optional features available. None are required for a fully functional VPN server. These optional features are described in greater detail in [here](docs/deploy-from-ansible.md).

That's it! You will get the message below when the server deployment process completes. Take note of the p12 (user certificate) password and the CA key in case you need them later, **they will only be displayed this time**.

You can now set up clients to connect to your VPN. Proceed to [Configure the VPN Clients](#configure-the-vpn-clients) below.

```
    "#                          Congratulations!                            #"
    "#                     Your Algo server is running.                     #"
    "#    Config files and certificates are in the ./configs/ directory.    #"
    "#              Go to https://whoer.net/ after connecting               #"
    "#        and ensure that all your traffic passes through the VPN.      #"
    "#                     Local DNS resolver 172.16.0.1                    #"
    "#        The p12 and SSH keys password for new users is XXXXXXXX       #"
    "#        The CA key password is XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX       #"
    "#      Shell access: ssh -i configs/algo.pem root@xxx.xxx.xx.xx        #"
```

TC1977:
Do we have to use `pip3` now as well?

davidemyers:
I *think* `python3 -m pip` is equivalent to `pip3`.

davidemyers:
One thing I've noticed with this branch is that when deploying to DigitalOcean the SSH key is no longer deleted from DO when the deployment finishes.

jackivanov:
@davidemyers This feature was introduced long time ago as a workaround for some unknown either Ansible or DigitalOcean bugs, IIRC. The new DigitalOcean modules seems to work pretty stable, and we don't need this anymore. We also don't delete the key in the others clouds. 

I thought of injecting the ssh key through the cloud-init, so we can avoid uploading it to the cloud database (already implemented for EC2, and a similar solution for GCE), but DigitalOcean sends an email with the root password if no ssh key specified for the instance. It's not a problem to remove the password form the root account, but it may confuse the users, what do you think?

davidemyers:
@jackivanov I think you should continue to prevent that password email from being sent even if it means the account fills up with old keys.

davidemyers:
I ran into a problem with Lightsail, running from 18.04 (so Python 3.6):
```
TASK [cloud-lightsail : Install requirements] **************************************************************************
changed: [localhost]

TASK [cloud-lightsail : set_fact] **************************************************************************************
ok: [localhost]
 [WARNING]: Found internal 'results' key in module return, renamed to 'ansible_module_results'.


TASK [cloud-lightsail : Get regions] ***********************************************************************************
ok: [localhost]

TASK [cloud-lightsail : Set facts about the regions] *******************************************************************
fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'results'\n\nThe error appears to be in '/home/dem/algo/roles/cloud-lightsail/tasks/prompts.yml':
line 33, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n  - name: Set facts about the regions\n    ^ here\n"}

```

jackivanov:
Thanks @davidemyers! Btw, can you send a PR to this branch with your changes, please?

davidemyers:
@jackivanov I'll send a PR. If there are no objections, in the PR I will also remove `deploy-from-fedora-workstation.md` and move the Linux IPsec section from the README into a separate file.

jackivanov:
Why to remove `deploy-from-fedora-workstation.md`? I think it's still relevant

davidemyers:
Fedora is now covered in my changes to the README above. I think the old documentation is unnecessarily complicated.

jackivanov:
ok, got it. I think there was a reason we included all those others packages to install, I wouldn't exclude them without proper testing.

davidemyers:
I've tested all of the instructions in my suggested changes to the README (Ubuntu 18.04 & 19.04, Debian 10, Fedora 30, and macOS 10.14.6 with Python from both Homebrew and the python.org package). But more testing is of course welcome.

All of my Linux tests were with server versions, as I assumed they were subsets of desktop versions. Also I only tested DigitalOcean, assuming the other providers would not require other system prerequisites.

jackivanov:
As @davidemyers [noted](https://github.com/trailofbits/algo/issues/1571#issuecomment-526698603), the DO module fails now. I'm not sure how soon the [PR](https://github.com/ansible/ansible/pull/61643) will be merged to the upstream, so I just put the fixed module locally

azdkj532:
> I'm not a Python expert, but I think [PEP 538](https://www.python.org/dev/peps/pep-0538/) explains the issue with generating the p12 password in Python 3. This works at the command line on 18.04, 19.04, and macOS:
> 
> ```
> openssl rand 8 | LANG=C python3 -c 'import sys,string; chars=string.ascii_letters + string.digits + "_@"; print("".join([chars[ord(c) % 64] for c in list(sys.stdin.read())]))'
> ```
> 
> I haven't tried it this way yet when invoked from Ansible.

Hi,
```
azdkj532@MacBookAir ~/Workspace/algo$ openssl rand 8 | LANG=C python3 -c 'import sys,string; chars=string.ascii_letters + string.digits + "_@"; print("".join([chars[ord(c) % 64] for c in list(sys.stdin.read())]))'

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/larry/.pyenv/versions/3.7.1/Python.framework/Versions/3.7/lib/python3.7/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x91 in position 0: invalid start byte
```
I found this code doesn't work on my MacBookAir with python3.7. 
In python3, I suggest to use `stdin.buffer` to read binary data from stdin, but this attribute is invalid in python2.

bhawkins:
I just installed from this branch on WSL, though I had to patch an Ansible script:

```
$ diff -Naur .env/lib/python3.5/site-packages/ansible/modules/cloud/digital_ocean/di
gital_ocean_sshkey.py{.orig,}
--- .env/lib/python3.5/site-packages/ansible/modules/cloud/digital_ocean/digital_ocean_sshkey.py.orig   2019-08-31 15:01:22.606761500 -0700
+++ .env/lib/python3.5/site-packages/ansible/modules/cloud/digital_ocean/digital_ocean_sshkey.py        2019-08-31 15:02:23.758253300 -0700
@@ -107,7 +107,7 @@
                 return json.loads(self.info["body"])
             return None
         try:
-            return json.loads(self.body)
+            return json.loads(self.body.decode())
         except ValueError:
             return None
```

jackivanov:
It appears that the minimum required version of python should be [3.6](https://docs.python.org/3/whatsnew/3.6.html#json), and seem like ubuntu 16.04 is incompatible now

davidemyers:
You should also remove the sentence that starts with "On Ubuntu 16.04 LTS ..." that I recently added to the README.

davidemyers:
This branch works again without `library/digital_ocean_droplet.py`.

docs/deploy-from-script-or-cloud-init-to-localhost.md
```diff
@@ -18,18 +18,30 @@ The command will prepare the environment and install AlgoVPN with the default pa
...
+- `METHOD`: which method of the deployment to use. Possible values are local and cloud. Default: cloud. The cloud method is intended to use in cloud-init deployments only. If you are not using cloud-init to deploy the server you have to use the local method.
+
 - `ONDEMAND_CELLULAR`: "Connect On Demand" when connected to cellular networks. Boolean. Default: false.
+
 - `ONDEMAND_WIFI`: "Connect On Demand" when connected to Wi-Fi. Default: false.
+
 - `ONDEMAND_WIFI_EXCLUDE`: List the names of any trusted Wi-Fi networks where macOS/iOS IPsec clients should not use "Connect On Demand". Comma-separated list.
-- `STORE_PKI`: To retain the PKI. (required to add users in the future, but less secure). Default: false.
+
+- `STORE_PKI: To retain the PKI. (required to add users in the future, but less secure). Default: false.
```
TC1977: missing a backtick here. Should be:

```
`STORE_PKI`
```

library/gcp_compute_location_info.py
```diff
@@ -0,0 +1,93 @@
...
+    if not module.params['scopes']:
+        module.params['scopes'] = ['https://www.googleapis.com/auth/compute']
+
+    items = fetch_list(module, collection(module), query_options(module.params['filters']))
+    if items.get('items'):
+        items = items.get('items')
+    else:
+        items = []
+    return_value = {'resources': items}
+    module.exit_json(**return_value)
```
btonic: Could this be simplified to the following?:
```python
module.exit_json(**dict(resources=items.get('items', [])))
```

library/gcp_compute_location_info.py
```diff
@@ -0,0 +1,93 @@
...
+    return "https://www.googleapis.com/compute/v1/projects/{project}/{scope}".format(**module.params)
+
+
+def fetch_list(module, link, query):
+    auth = GcpSession(module, 'compute')
+    response = auth.get(link, params={'filter': query})
+    return return_if_object(module, response)
+
+
+def query_options(filters):
```
btonic: I think we could make this a little more concise with list comprehensions:
```python
def query_options(filters):
    if not filters:
        return ''
    elif len(filters) == 1:
        return filters[0]
    else:
        return ' '.join(
            filter if (filter.startswith('(') and filter.endswith(')'))
                   else f"({filter})"
            for filter in filters
        )
```

btonic: Actually, the first two cases are implicitly handled if we turn it into a generator:
```python
def query_options(filters):
    return ' '.join(
        filter if (filter.startswith('(') and filter.endswith(')'))
               else f"({filter})"
        for filter in filters
    )
```

library/gcp_compute_location_info.py
```diff
@@ -0,0 +1,93 @@
...
+                queries.append("(%s)" % ''.join(f))
+            else:
+                queries.append(f)
+
+        return ' '.join(queries)
+
+
+def return_if_object(module, response):
+    # If not found, return nothing.
+    if response.status_code == 404:
```
btonic: we can merge these together:
```python
if response.status_code in (404, 204,):
    return None
```

Issue #989, opened 2018-06-04
EC2 AWS instance volumes aren't encrypted


TC1977:
### OS / Environment (where do you run Algo on)

```
Darwin Mycomputername.local 17.5.0 Darwin Kernel Version 17.5.0: Fri Apr 13 19:32:32 PDT 2018; root:xnu-4570.51.2~1/RELEASE_X86_64 x86_64
```

### Cloud Provider (where do you deploy Algo to)

```
AWS EC2
```

### Summary of the problem
I'm attempting to help someone on Gitter and just installed an instance on AWS EC2. Looking at the console, the volume isn't encrypted. Wasn't this encrypted previously?

### Steps to reproduce the behavior

1. Install Algo onto EC2 AWS
2. Check the console, review the volume info (or install awscli)

### Full log

```
(env) Mycomputername:algo-master-2 admin$ ./algo

  What provider would you like to use?
    1. DigitalOcean
    2. Amazon EC2
    3. Microsoft Azure
    4. Google Compute Engine
    5. Scaleway
    6. OpenStack (DreamCompute optimised)
    7. Install to existing Ubuntu 16.04 server (Advanced)

Enter the number of your desired provider
: 2

Enter your aws_access_key (http://docs.aws.amazon.com/general/latest/gr/managing-aws-access-keys.html)
Note: Make sure to use an IAM user with an acceptable policy attached (see https://github.com/trailofbits/algo/blob/master/docs/deploy-from-ansible.md).
[pasted values will not be displayed]
[AKIA...]: 

Enter your aws_secret_key (http://docs.aws.amazon.com/general/latest/gr/managing-aws-access-keys.html)
[pasted values will not be displayed]
[ABCD...]: 

Name the vpn server:
[algo]: algo1


  What region should the server be located in?
    1.   us-east-1           US East (N. Virginia)
    2.   us-east-2           US East (Ohio)
    3.   us-west-1           US West (N. California)
    4.   us-west-2           US West (Oregon)
    5.   ca-central-1        Canada (Central)
    6.   eu-central-1        EU (Frankfurt)
    7.   eu-west-1           EU (Ireland)
    8.   eu-west-2           EU (London)
    9.   eu-west-3           EU (Paris)
    10.  ap-northeast-1      Asia Pacific (Tokyo)
    11.  ap-northeast-2      Asia Pacific (Seoul)
    12.  ap-northeast-3      Asia Pacific (Osaka-Local)
[...]

PR #1556, opened 2019-08-19
EC2: Enable EBS single step encryption by default

jackivanov:
## Description
- Disable the ami copy task that takes too much time to make an encrypted volume
- Enable EC2 EBS single step encryption by default

## Motivation and Context
As @TC1977 noted https://github.com/trailofbits/algo/issues/989#issuecomment-522084757, it is possible now to enable EBS encryption in one go without waiting too much time. Also, it makes sense now to enable it by default.

## How Has This Been Tested?
Deployed to EC2 with and without encryption.

## Types of changes

## Checklist:


TC1977:
Tested successfully on EC2 us-east-1 using non-interactive Ansible deploy, both with "encrypted=true" and "encrypted=false".

With default settings of everything false, I have a run time of about 8:16 with -vvv and encrypted=true, and _8:39_ with -vvv and encrypted=false. I have no idea why the encrypted run was faster (maybe because I was deleting the CloudFormation at the same time for the unencrypted run?), but at least there isn't a 6-minute performance hit with encryption like before.

TC1977:
Ok, that explains why the encrypted run was faster - both runs were encrypted. 😆 

Running the newest commit with "encrypted=true" vs false and no -vvv gives run times of 8:38 with encryption vs. 8:28 without encryption.

Issue #1588, opened 2019-09-25
Middle East, China and Turkey Disclaimer


shadowsaw:
This is a general Disclaimer for everyone trying to install Algo in censorship prone areas that the **Wireguard protocol is often blocked**. So performance issues are to be expected. Try manual ikev2 or other VPN installations.

shadowsaw:
The VPN behaves similar to the problem described in #1586

TarekSalama:
How is manual ikev2 different from the one installed by algo?

TarekSalama:
@shadowsaw

shadowsaw:
Algo installs two protocols only. Wireguard and ikev2. You can read up more abt the differences between the two.
In the config folder, at the end of successful installation, you'll find 3 sub-folders. 
The "Apple" folder generates an iOS-compatible profile(s) which makes use of native Ikev2/IPsec capabilities of iPhones.
The "wireguard" folder generates qr codes. It uses specifically the wireguard protocol only.
Lastly the "manual" folder, has client certificates for each user as well as one certificate-authority certificate. Import those into Strongswan or any other app of your choice, to use ikev2 protocol on Android.

JACKHAHA363:
#1743 After installation I only have a `.p12` file but the guide said I need `.crt` and `.key` to manually install the ipsec. So how should I generate these two files from `.p12`?

shadowsaw:
You have to do your install of Algo. Key is displayed at the end of installation only and never again. So make sure to take note of it then

AhmedSamirAhmed:
Same for me (Egypt). Suddenly most of clients (Same ISP) stopped working after working fine for around 5 months.
But i have seen a strange behavior which is if i started a working VPN (Hotspot shield for ex) on my windows machine then go to my debian vm installed on vmware workstation (Using Natting network setting) and started wireguard client the client connects successfully and started working fine even i've stopped the windows VPN client.
So I'm asking for any tips to bypass the handshaking dropping issue i got when using wireguard on windows if there is any

yehia2amer:
Could this method be automated ?

https://www.ntkernel.com/how-to-bypass-egypts-wireguard-ban/

dguido:
Hi @shadowsaw,

Thank you for reporting this WireGuard issue. Since you reported this, WireGuard has reached stable 1.0.0 release (Spring 2020) and has undergone substantial security review.

WireGuard is now:
- Officially stable and production-ready
- Included in the Linux kernel
[...]

Issue #1586, opened 2019-09-16
Fortiguard identifies and blocks Wireguard


TarekSalama:
So I was lucky enough to be able to test Wireguard with Fortiguard installed on the network.
Fortiguard was able to completely disable Wireguard.

I did notice however that it sometimes identified it as a proxy (Wireguard) and sometimes it identified it as an AWS app (where I host it). I did notice also that it sometimes worked and sometimes it worked only after reconnecting to the network. I couldn't find an exact pattern.

Is there a way to completely hide it from getting identified by Fortiguard? Any idea why it was sometimes identified as a proxy/AWS app?

TC1977:
What if you change the Wireguard port to 1194 (Open VPN) or 443 (https)? Or something completely random?

TarekSalama:
I am not really experienced with networking but is it possible to change it using Algo?
I know I can probably change it in the server script configuration file but would I need to change anything in the device config files that get auto generated?

dguido:
Hi @TarekSalama,

Thank you for reporting this WireGuard issue. Since you reported this, WireGuard has reached stable 1.0.0 release (Spring 2020) and has undergone substantial security review.

WireGuard is now:
- Officially stable and production-ready
- Included in the Linux kernel
- Actively maintained with regular updates
- The recommended VPN protocol for Algo

The issues you experienced were likely related to the pre-1.0 development versions. These have been resolved in the stable releases.

Since WireGuard has matured significantly and this issue relates to early development versions, I'm closing it. If you encounter any issues with current versions of WireGuard and Algo, please feel free to open a new issue.

Thank you for your patience and for using Algo!

PR #1594, opened 2019-10-01
Allow WireGuard to listen on port 53

### LINKED ISSUE #1593: Support for using port 53 for Wireguard

As Wireguard is operating on UDP, it is quite common on corporate and some public networks to be impossible to reach the VPN server as most UDP ports are blocked. A common tactic is then to setup Wireguard to listen on port 53, usually used by DNS, as it is very rarely filtered. This is a similar approach as using TCP443 for OpenVPN. However, if you try to put 53 as the Wireguard port in config.cfg, the deployment breaks when it tries to start the Wireguard service (wg-quick@wg0) as the port is already used by Systemctl resolved/dnscrypt-proxy.

**Describe the solution you'd like**

Could the local dns resolver /dnscrypt proxy be listening on another IP? Such as an ip within the wireguard default subnet like 10.19.49.53? Or 172.16.10.53?

**Describe alternatives you've considered**

I've tried changing the dns resolver to listen on port 5353 but I could not find a way to edit Wireguard client profiles to change the DNS listening port.

I've tried setting up the dns on another IP as suggested in the solution as well but to no avail, however, I'm pretty sure the issue lies between keyboard and chair and not into the solution itself ;).

davidemyers:
## Description

This change allows the special case of setting `wireguard_port: 53` in `config.cfg` and is based on an approach suggested on the [WireGuard mailing list](https://lists.zx2c4.com/pipermail/wireguard/2018-June/003013.html). 

## Motivation and Context

Some networks block most outgoing UDP traffic, but port 53 might still be allowed for DNS.

Closes #1593.

## How Has This Been Tested?

Deployed to DigitalOcean both with and without `wireguard_port: 53`. Also tested connecting using a DigitalOcean Floating IP, where traffic comes in on the main network interface but to a secondary address. In all cases the client was iOS 13.1.2.

Also tested on Lightsail with `wireguard_port: 53`.

More testing should be done. Please help test this, fellow Algo users.

This should also be reviewed for how it uses `iptables`.

## Types of changes

## Checklist:

### DIFF
--- a/roles/common/templates/rules.v4.j2
+++ b/roles/common/templates/rules.v4.j2
@@ -1,5 +1,5 @@
 {% set subnets = ([strongswan_network] if ipsec_enabled else []) + ([wireguard_network_ipv4] if wireguard_enabled else []) %}
-{% set ports = (['500', '4500'] if ipsec_enabled else []) + ([wireguard_port] if wireguard_enabled else []) + ([wireguard_port_alt] if wireguard_enabled and wireguard_port|int == 53 else []) %}
+{% set ports = (['500', '4500'] if ipsec_enabled else []) + ([wireguard_port] if wireguard_enabled else []) + ([wireguard_port_actual] if wireguard_enabled and wireguard_port|int == wireguard_port_avoid|int else []) %}
 
 #### The mangle table
 # This table allows us to modify packet headers
@@ -29,9 +29,10 @@ COMMIT
 :PREROUTING ACCEPT [0:0]
 :POSTROUTING ACCEPT [0:0]
 
-{% if wireguard_enabled and wireguard_port|int == 53 %}
-# Handle the special case of allowing access to WireGuard over port 53
--A PREROUTING --in-interface {{ ansible_default_ipv4['interface'] }} -p udp --dport 53 -j REDIRECT --to-port {{ wireguard_port_alt }}
+{% if wireguard_enabled and wireguard_port|int == wireguard_port_avoid|int %}
+# Handle the special case of allowing access to WireGuard over an already used
+# port like 53
+-A PREROUTING --in-interface {{ ansible_default_ipv4['interface'] }} -p udp --dport {{ wireguard_port_avoid }} -j REDIRECT --to-port {{ wireguard_port_actual }}
 {% endif %}
 # Allow traffic from the VPN network to the outside world, and replies
 -A POSTROUTING -s {{ subnets|join(',') }} -m policy --pol none --dir out -j MASQUERADE
--- a/roles/common/templates/rules.v4.j2
+++ b/roles/common/templates/rules.v4.j2
@@ -1,5 +1,5 @@
 {% set subnets = ([strongswan_network] if ipsec_enabled else []) + ([wireguard_network_ipv4] if wireguard_enabled else []) %}
-{% set ports = (['500', '4500'] if ipsec_enabled else []) + ([wireguard_port] if wireguard_enabled else []) %}
+{% set ports = (['500', '4500'] if ipsec_enabled else []) + ([wireguard_port] if wireguard_enabled else []) + ([wireguard_port_alt] if wireguard_enabled and wireguard_port|int == 53 else []) %}
 
 #### The mangle table
 # This table allows us to modify packet headers
@@ -29,6 +29,10 @@ COMMIT
 :PREROUTING ACCEPT [0:0]
 :POSTROUTING ACCEPT [0:0]
 
+{% if wireguard_enabled and wireguard_port|int == 53 %}
+# Handle the special case of allowing access to WireGuard over port 53
+-A PREROUTING --in-interface {{ ansible_default_ipv4['interface'] }} -p udp --dport 53 -j REDIRECT --to-port {{ wireguard_port_alt }}
+{% endif %}
 # Allow traffic from the VPN network to the outside world, and replies
 -A POSTROUTING -s {{ subnets|join(',') }} -m policy --pol none --dir out -j MASQUERADE
 
--- a/config.cfg
+++ b/config.cfg
@@ -32,6 +32,9 @@ strongswan_network: 10.19.48.0/24
 strongswan_network_ipv6: 'fd9d:bc11:4020::/48'
 
 # Deploy WireGuard
+# WireGuard will listen on 51820/UDP. You might need to change to another port
+# if your network blocks this one. Be aware that 53/UDP (DNS) is blocked on some
+# mobile data networks.
 wireguard_enabled: true
 wireguard_port: 51820
 # If you're behind NAT or a firewall and you want to receive incoming connections long after network traffic has gone silent.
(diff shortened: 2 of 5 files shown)


davidemyers:
The Travis failure doesn't appear to be related to my changes.

TC1977:
Wonder if this would help solve #1586 and #1588.

davidemyers:
I've found an issue that I don't believe is specific to this PR but should be noted as it is likely a showstopper for many.

Using port 53 with WireGuard does not appear to work at all over Verizon Wireless LTE (using an iPhone, anyway). I don't have a way to test other carrier networks.

roles/common/templates/rules.v4.j2
```diff
@@ -29,6 +29,10 @@ COMMIT
 :PREROUTING ACCEPT [0:0]
 :POSTROUTING ACCEPT [0:0]
 
+{% if wireguard_enabled and wireguard_port|int == 53 %}
+# Handle the special case of allowing access to WireGuard over port 53
```
jackivanov: Why would not make port 53 also configurable? We can keep `wireguard_port: 51820` and the new variable `wireguard_port_alt: 53`, where `_alt` means an alternative port which is configured by iptables?

davidemyers: I'm not sure I understand your suggestion. I didn't want to change the meaning of `wireguard_port`, or the code to generate client config files and to open provider firewall ports would also have to change. I also intended to make this change transparent to users.

I felt hardcoding `53` was OK because there are already several hardcoded port numbers in the `iptables` templates.

There's probably a better name for `wireguard_port_alt`, which is where WireGuard is told to *really* listen if the user happens to set `wireguard_port: 53` and thereby trigger this code. I was trying to keep it short.

If I was better at Jinja2 I might have tested `wireguard_port` against an array or list of numbers, not just `53`. For example, setting `wireguard_port` to `68` or `546` might also fail without this `iptables` trick, depending on the provider. Actually, this trick might not work for `68` on EC2.

jackivanov: My point was about not hardcoding 53 port and make it configurable

davidemyers: We could do that. I think it needs to be a new variable. Should it be in `config.cfg` or the WireGuard defaults where `_alt` is now?

davidemyers: I changed to using a variable for the port. I think I'll also update the comments in `config.cfg` to mention possible issues with mobile data.

TC1977:
I'm gonna try to spin up a server later today to test this out on AT&T LTE network.

Wouldn't this also be a useful feature for IPsec users as well? One big appeal of OpenVPN is to select the outgoing port at will. You could set the outgoing port to whatever you want in config.cfg, set the open cloud/local firewall ports accordingly, and use iptables to redirect it to UDP/500 and 4500 on the Algo server. Might be outside the scope of this PR, though.

davidemyers:
I don't believe there's a way of tell most IPsec clients to use alternate ports.

I'm curious to know if this works over AT&T.

TC1977:
Ok, so here's what I did:
Created an Ubuntu 19.04 droplet on DO
Used `git clone` and `git checkout wg53` to switch to this PR
Ran through steps 3-5 of the install instructions
Changed wireguard_port in config.cfg to 53
Ran ./algo and chose local install, adblocking, save CA key, no other options.
Created a DO firewall to allow UDP/53 in and TCP/22 only.
Connected using the laptop.conf on MBP on LAN to server, web browsing and adblocking work.
Connected using the phone.png on iPhone on LAN to server, web browsing and adblocking work.
Switched to LTE - connection drops instantly, app log says "Handshake did not complete after 5 seconds, retrying"
Switched back to Wi-Fi - web browsing and adblocking work again.

davidemyers:
Thanks for the testing, that's the same behavior I found. It also fails on an iPad when using the iPhone's Personal Hotspot (hotspot traffic is sent "in the clear", not over the phone's WireGuard tunnel, which is nice because it avoids double tunneling and the resulting MTU issues).

So this PR is a lot less useful than I had hoped it would be.

TC1977:
It's not like I have a better idea, but can we have more descriptive names than `wireguard_port_actual` and `wireguard_port_avoid`? I would have no idea what those mean.

davidemyers:
I'm open to suggestions. Where they come from is WireGuard must "avoid" listening on port 53 and instead "actually" listen somewhere else.

Since those variables aren't in `config.cfg` and won't be seen by most users I'm not sure the naming is really that important.

TC1977:
So if someone sets `wireguard_port` in `config.cfg` to port 53 (or whatever you've set `wireguard_port_avoid` to), iptables will silently redirect traffic to the standard port of 51820 instead. And Wireguard won't actually be listening on port 53, but on port 51820. Is that correct?

It might be good to document this behavior in `config.cfg`, as I don't think anyone would expect it. I'd also suggest adding `wireguard_port_avoid` as a potential variable in the Ansible docs. (I'm still trying to think of a more intuitive name.)

davidemyers:
> So if someone sets `wireguard_port` in `config.cfg` to port 53 (or whatever you've set `wireguard_port_avoid` to), iptables will silently redirect traffic to the standard port of 51820 instead. And Wireguard won't actually be listening on port 53, but on port 51820. Is that correct?

Yes, that's the recommended approach according to the mailing list message I linked to above.

> It might be good to document this behavior in `config.cfg`, as I don't think anyone would expect it.

My intention was to make this transparent to the user so that setting port 53 in `config.cfg` "just works". We now know that mobile data providers have other ideas, but I don't see why a user needs to know that `iptables` trickery is going on.

TC1977:
It does make sense to make this transparent to the user, but I see two issues:

- Algo obviously serves a lot of the "prosumer" or outright pros who have an idea what they want to accomplish, but may not have the time/knowledge to go hunting around for iptables rules and WireGuard mailing list suggestions. Therefore it won't prevent future issues opened by people who want to have WireGuard listen on some other reserved port.
- The more places you put these options, the harder it will be to maintain these features as time goes on and the underlying software gets upgraded.

I'd suggest renaming it `wireguard_port_avoid` to `wireguard_port_masquerade` (sounds iptables-ish 😄 )  and putting it in `config.cfg` with appropriate documentation. There's already precedent for putting iptables rules mods with `block_smb` and `BetweenClients_DROP` in `config.cfg`.

davidemyers:
The term "masquerade" has a particular meaning in `iptables` and I think using it here could be misleading.

The problem with promoting the variable to `config.cfg` is that it implies this feature is more generally useful than it really is. For this approach to work, nothing must be listening to the conflicting port on any IP address assigned to the primary network interface, meaning it won't work for most port conflicts a user is likely to encounter.

In the case of port 53, on an AlgoVPN `systemd-resolved` (by default) and `dnscrypt-proxy` (as configured by Jack) listen on private addresses, not on the primary network interface. However I can't think of another port likely to be used on a server where only private addresses would be used, and that a user would also want to use for WireGuard.

TC1977:
> In the case of port 53, on an AlgoVPN `systemd-resolved` (by default) and `dnscrypt-proxy` (as configured by Jack) listen on private addresses, not on the primary network interface. However I can't think of another port likely to be used on a server where only private addresses would be used, and that a user would also want to use for WireGuard.

Yeah, neither could I. Well, I have no further objections then.

msuiche:
I just want to say that his is a beautiful pull request.

PR #1595, opened 2019-10-02
Troubleshooting docs addition for CA certificate failure with MacPorts

jonstewart:
Added a section to the troubleshooting doc for an SSL certificate validation error that occurs when using MacPorts instead of Homebrew.

The openssl port in MacPorts does not include a valid CA certificate. This is a long-standing issue, but it can be fixed by installing the `curl-ca-bundle` port. See https://andatche.com/articles/2012/02/fixing-ssl-ca-certificates-with-openssl-from-macports/.

I tried installing `algo` last night on a Mac where the ports had not been heavily used and ran into this error. `sudo port install curl-ca-bundle` was all that was needed to get past it.

Thanks for the great tool!

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -122,13 +122,10 @@ You did not finish step 4 in the installation instructions, "[Install Algo's rem
 
 You received a message like this:
 ```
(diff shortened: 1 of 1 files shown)


jonstewart:
Thanks!

PR #1605, opened 2019-10-07
Alternative Ingress IP

### LINKED ISSUE #1047: Separate egress IP address
I am in the process of adding a second nic on the algo VM with the intention of sending outbound traffic from the VM through a different external IP address than the one I'm establishing the VPN connection with.  The goal here is to disassociate the two so as to make it more difficult for an adversary to correlate activity, where they have knowledge that I have a VPN tunnel established to an IP address, and may be able to see traffic leaving the provider's network or arriving at servers originating from that IP address.

It's unclear to me whether this feature fits the algo feature/anti-feature criteria, since this adds complexity for a benefit that most algo users may not care about, so I'm not sure whether I should aim to contribute my changes upstream.  Thoughts?

jackivanov:
## Description
This PR adds functionality to provision and configure a separate ingress IP to split up incoming and outgoing traffic. The PR is for DigitalOcean only for now.
Because of a [bug](https://github.com/ansible/ansible/issues/63175) in Ansible, we're again forced to modify the module locally.

## Motivation and Context
Fixes #1047

## How Has This Been Tested?
1. Deployed to DigitalOcean
2. Connected via IKEv2 and WireGuard
3. Ensured that the endpoint for both IKEv2 and WireGuard was set to the Floating IP
4. Went to ipleak.net to ensure the ip shown was not the one from the step 3

## Types of changes

## Checklist:

## TODO:
- [x] Short but efficient description comment [here](https://github.com/trailofbits/algo/pull/1605/files#diff-3931db100a156fcf1aac55b530183ea8R46)
- will be addressed as a separate issue: As we now provision multiple resources without a shared control mechanism (like CloudFormation), we have to create a mechanism to clean up the environment, delete the server and the floating ip by running `./algo destroy` which rises cloud specific prompts and displays the resources available for deletion.

Any suggestions welcome!


jackivanov:
cc @dnesting @davidemyers

davidemyers:
I suggest a name other than `static_ip` for the variable in `config.cfg` as I think that name will confuse people (the IPs for most (all?) providers other than Lightsail can be thought of as static). Perhaps `second_ip` or `ingress_ip`.

davidemyers:
I also suggest that the `configs/` subdirectory be named after the primary IP address rather than the secondary ingress IP.

config.cfg
```diff
@@ -43,6 +43,10 @@ wireguard_PersistentKeepalive: 0
 wireguard_network_ipv4: 10.19.49.0/24
 wireguard_network_ipv6: fd9d:bc11:4021::/48
 
+# Used to segregate the incoming traffic to a separate IP
+# Available for the following cloud providers: DigitalOcean
+static_ip: true
```
jackivanov: @davidemyers How do you find `alternative_ingress_ip` suitable for this?

davidemyers: Yes, thanks.

jackivanov:
Alright, it's ready for more tests

davidemyers:
This breaks IPv6 for me (tested via http://ipv6-test.com/ and https://browserleaks.com/ip). I don't see the alternate IPv6 address on `ens3` even after a reboot.

Deployed from Ubuntu 19.10 using the alternate Python 3.8 and tested using WireGuard on iOS.

jackivanov:
OK, seems we have to use netplan. @davidemyers I think it's fixed now, works on my end

davidemyers:
That works for me, thanks.

Should the secondary IPv6 address be /64 like the primary? I tried it that way and it seems to work.

davidemyers:
The docs should mention that when using this feature with DigitalOcean, after manually deleting a Droplet you need to also delete the Floating IP or you'll get charged for it.

jackivanov:
> Should the secondary IPv6 address be /64 like the primary? I tried it that way and it seems to work.

It will work for sure, but DigitalOcean actually [gives](https://www.digitalocean.com/docs/networking/ipv6/how-to/configure-additional-addresses/) you a `/124`, but not the whole `/64`, so [we take](https://github.com/trailofbits/algo/pull/1605/files#diff-606b407a34fd6cfde84c2c1a3190044cR5-R6) a random address from the `/124` subnet

benturner:
Hm, this doesn't appear to do anything for the IPv4 traffic?

jackivanov:
@benturner It's disabled by default at the moment. You need to turn it on in config.cfg `alternative_ingress_ip: false`. Can you try that?

benturner:
Yep, I did. It properly changed my `rules.v6` file but the `rules.v4` file didn't get the `SNAT` option added to it.

jackivanov:
@benturner It doesn't modify rules.v4, it just sets the destination IP for the vpn connection as the Floating IP. 

But I think it might be not good way in case we want to add more providers

jackivanov:
I've refactored the code, so we use the simplest schema with SNAT

jackivanov:
@davidemyers could you check out this one please?

davidemyers:
Works for me, tested from WireGuard on iOS.

PR #1608, opened 2019-10-09
fixed code to work with python3.7

0xmilkmix:
## Description
I removed a fixed path in the role that used `python2.7`. Also saw that a `block` was not properly defined causing the `prompts` module not to be properly called.

## Motivation and Context
I saw that the module was not working properly after `python3` support changes

I have a patch to use `cs_zone_info` instead of homemade `cloudstack_zones.py` but it requires `ansible2.9` so I'll wait for release and perform more test for global support of this release.

## How Has This Been Tested?
created instance on cloudstack service provided by Exoscale

## Types of changes
## Checklist:

### DIFF
--- a/roles/cloud-cloudstack/tasks/main.yml
+++ b/roles/cloud-cloudstack/tasks/main.yml
@@ -1,70 +1,64 @@
 ---
+- name: Build python virtual environment
+  import_tasks: venv.yml
+
+- name: Include prompts
+  import_tasks: prompts.yml
+  
 - block:
-    - name: Build python virtual environment
-      import_tasks: venv.yml
+    - set_fact:
+        algo_region: >-
+          {% if region is defined %}{{ region }}
+          {%- elif _algo_region.user_input is defined and _algo_region.user_input | length > 0 %}{{ cs_zones[_algo_region.user_input | int -1 ]['name'] }}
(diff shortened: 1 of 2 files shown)


roles/cloud-cloudstack/tasks/main.yml
```diff
@@ -1,70 +1,64 @@
...
-            {%- else %}{{ cs_zones[default_zone | int - 1]['name'] }}{% endif %}
-
-      - name: Security group created
-        cs_securitygroup:
-          name: "{{ algo_server_name }}-security_group"
-          description: AlgoVPN security group
-        register: cs_security_group
+- name: Include prompts
+  import_tasks: prompts.yml
+
```
davidemyers: Trailing whitespace here is causing the build to fail.

roles/cloud-cloudstack/tasks/main.yml
```diff
@@ -1,70 +1,64 @@
...
-        tags: always
-      - fail:
-        tags: always
+    - set_fact:
+        cloud_instance_ip: "{{ cs_server.default_ip }}"
+        ansible_ssh_user: ubuntu
+  environment:
+    CLOUDSTACK_CONFIG: "{{ algo_cs_config }}"
+    CLOUDSTACK_REGION: "{{ algo_cs_region }}"
+
```
davidemyers: Trailing whitespace here is causing the build to fail.

Issue #196, opened 2016-12-31
Webapp that builds Algo VPNs


dguido:
Form fields for all the info we need, click create, it spits out the configs at you and downloads and tells you the server info.

Like this: https://www.tinfoilsecurity.com/vpn/new

MiWCryptAnalytics:
Hi Dan,
I have been thinking about implementing something similar and discussed with you at EH.
I would like to do DO and AWS implementations, as have been doing lots of AWS InfAsCode 
work lately.

Ill start working on this now, you assign this issue to me?

dguido:
Sure! Here's a good template to start from for a UI:
https://www.tinfoilsecurity.com/vpn/new

MiWCryptAnalytics:
My plan is to use two separate repos, one for the flask app container and another for the Infrastructure as Code scripts that the flask application will use to provision with DO and AWS.
Having both repos open source will hopefully increase trust in the app not to do nefarious things with other peoples tokens.

The app will pull from this repo and run the build operations for each VPN setup in an ephemeral container. Need to do something about trust and/or liability with handling others API tokens also.

In spirit of Tinfoil security launching their service for the Rio Olympics, on 13 April 2017, Australia's mandatory data retention scheme kicked in:
https://iapp.org/news/a/australias-mandatory-data-retention-scheme-takes-effect-april-13/

Thanks for collab invite.

dguido:
Sounds fine. Let's start with the Flask app. Then somehow can just start it locally to get a GUI of sorts for the script. That will be an easy improvement.

MiWCryptAnalytics:
Yo dan, Ive made a start on this, im working on it out of here. 
https://github.com/MiWCryptAnalytics/algo-flask/

It doesn't do much yet and is pretty basic, 
but it does automatically generate the forms from the script via regex.
Is this crazy? Interested in your opinion before i do too much more with that

https://github.com/MiWCryptAnalytics/algo-flask/blob/master/generate.py

Ill make it look nicer in the next few days

dguido:
Nice! You should submit a WIP PR. Putting it in an `app` directory sounds appropriate?

dguido:
I think a better approach might be to have the webapp re-implement what is currently in the bash script. I don't think you need to do any fancy parsing of the shell script to generate the forms. We only ask about 5 or so questions. It's easy enough to hardcore a few toggles for each one.

MiWCryptAnalytics:
Your right, its too crazy. 

[...]

PR #1616, opened 2019-10-15
[DRAFT] Simple webapp for config

summerisgone:
Simple webapp for #196

## Description
Web server introduces two new requirements: pyyaml + aiohttp.
To launch it run `python app/server.py` and open [http://localhost:8080/](http://localhost:8080/). Server will show interactive forms for users configuration and will allow to proceed with digitalocean (only yet) setup.

Server launch:
Screenshot 2019-10-15 at 01 40 49

Custom fields for DigitalOcean:
image

Console output to track progress:
Screenshot 2019-10-15 at 09 34 11

## Motivation and Context
Your project helped me out when I had issues with ISP country-wide. I want to give back.

## How Has This Been Tested?
The change now is in PoC stage, I want to sync with project maintainers that work is done accordingly their vision of the project.

## Types of changes

## Checklist:

- ✅ I have read the **CONTRIBUTING** document.

- 🤷‍♂  My code follows the code style of this 

- [ ]  My change requires a change to the documentation.

## Open questions:
- I don't see clear way to keep only one source of truth for prompts. It is possible to sync web forms for existing setup, but will require further maintenance for new prompts
- I'm not sure about extra dependencies my project introduces. But I also think websocket for interactive console and pyyaml for parsing config files are nice things to have
- Am I correct that that form intended be used for hosted setup as well? That will require additional things to consider


requirements.txt
```diff
@@ -1,2 +1,5 @@
 ansible==2.8.3
 netaddr
+PyYAML==5.1.2
+aiodns==2.0.0
+aiohttp==3.6.2
```
jackivanov: I would keep the web requirements in a separate file instead

summerisgone: I thought about installing main requirements from web application in the background. Depends on what do you expect from that UI

app/server.py
```diff
@@ -0,0 +1,101 @@
...
+    ws = web.WebSocketResponse()
+    await ws.prepare(request)
+
+    async for msg in ws:
+        if msg.type == aiohttp.WSMsgType.TEXT:
+            if msg.data == 'close':
+                await ws.close()
+            else:
+                p = await asyncio.create_subprocess_shell(
+                    msg.data,
```
disconnect3d: This spawns a process based on data received from an unauthenticated websocket.

Which basically means... an attacker can prepare a website that would connect to localhost websocket (`new WebSocket("ws://127.0.0.1:8080/ws");`) and then send proper payload to execute arbitrary code on victim's machine.

It would be much better not to spawn any processes at all and if it is really needed, it must be authenticated.

app/server.py
```diff
@@ -0,0 +1,101 @@
...
+            'Authorization': 'Bearer {0}'.format(data['token']),
+        }
+        async with session.get(url, headers=headers) as r:
+            json_body = await r.json()
+            return web.json_response(json_body)
+
+app = web.Application()
+app.router.add_get('/ws', websocket_handler)
+app.router.add_get('/', handle_index)
+app.router.add_routes(routes)
```
disconnect3d: Is there any reason why some routes are added via decorating endpoints `@routes.post(...)` and then `app.router.add_routes(routes)` and some via `app.router.add_get(...)`?

We should be consistent and use one or another (ofc if possible).

app/server.py
disconnect3d: See also:
- https://medium.com/@sharan.panegav/account-takeover-using-cross-site-websocket-hijacking-cswh-99cf9cea6c50
- https://security.stackexchange.com/questions/76816/preventing-csrf-attacks-against-websocket-communications

summerisgone: @disconnect3d does Ansible provides API for launching playbooks other than from command line?

jackivanov: @summerisgone Yes, take a look at this https://docs.ansible.com/ansible/latest/dev_guide/developing_api.html

app/server.py
summerisgone: OK, will clean that up

app/server.py
summerisgone: Thanks, I probably missed that API from the first time. Will check it out!

summerisgone: @jackivanov I'm not good at ansible, could you help me or give an advice?
I'm able to run playbook from this snippet:
```python
PlaybookCLI(['ansible-playbook', 'main.yml']).run()
```
But in that case I can't display feedback, which is valuable, since installation process takes a long time and may exit with error. I'm looking for snippet which allows to set up custom `stdout_callback` and play scenario from yaml file, like above.

jackivanov: @summerisgone You don't seem to be running ansible via the API.  You can find a working example by the link I sent above

summerisgone: @jackivanov example you provided doesn't allow to load playbooks

jackivanov: I'm not well familiar with Ansible API, but as far I recall, there is a class called `PlaybookExecutor` that can be used. See [here](https://github.com/ansible/ansible/blob/stable-2.8/lib/ansible/executor/playbook_executor.py) and [here](https://github.com/ansible/ansible/blob/stable-2.8/test/units/executor/test_playbook_executor.py)

app/playbook.py
```diff
@@ -0,0 +1,195 @@
...
+                            taskmsg += _process_block(block)
+
+                        if context.CLIARGS['listtags']:
+                            cur_tags = list(mytags.union(all_tags))
+                            cur_tags.sort()
+                            taskmsg += "      TASK TAGS: [%s]\n" % ', '.join(cur_tags)
+
+                        display.display(taskmsg)
+
+        return inventory.groups['vpn-host'].hosts[0].vars
```
summerisgone: Here are changes from the original file:
https://github.com/ansible/ansible/blob/devel/lib/ansible/cli/playbook.py#L193-L195

I made assumption that playbook will provision only one host

summerisgone:
Updated description with fresh demo

summerisgone:
Guys, I need help with MS Azure provider, it doesn't allow to create account from my country! 🤯

app/server.py
summerisgone: @jackivanov or @disconnect3d could you please take another look on current approach of executing playbooks from web?

summerisgone:
Have issues with Hetzner as well: they disabled my account without description. Reaching out their support

SonyViao (approved):

summerisgone:
Rebased on recent master

summerisgone:
> Would the web interface also work on windows? I am interesting in progressing Algo VPN both as a project for school and to gain more insight into the programming world.

As long as Windows has modern browser (Edge, Chrome, Firefox) - web would work

anragab:
Would it make sense to have this UI in a separate repository?

summerisgone:
> Would it make sense to have this UI in a separate repository?

This branch haven't been updated for a long time, so it would require the rebase. Algo is focused on CLI interface and supposed to ask those questions on its side, see https://github.com/trailofbits/algo/issues/196. 

I don't have resources atm to support the codebase as a separate project, feel free to take it over )

dguido:
I'm going to pick this up myself some time soon. I still have a lot of other support requests to get through first though. I'm making slow progress, thanks to Claude Code.

dguido:
Ok welp, it's been 6 years but I finally got around to building this feature! Thank you @summerisgone. I'm going to work off my branch here and close this PR: https://github.com/trailofbits/algo/pull/14916

summerisgone:
If you need any help or review, don't mind to tag me!

PR #1652, opened 2019-12-07
Remove some commas from input.yml

davidemyers:
## Description

I previously added some commas to the local install prompt in `input.yml` as part of #1630, but it turns out the commas cause the prompt to be truncated.

## Motivation and Context

The prompt is wrong.

## How Has This Been Tested?

Initiated a local install on a new Droplet and observed the proper prompt.

## Types of changes

## Checklist:


TC1977:
What about just escaping the commas out with a quote or something?

davidemyers:
I'm not sure what the correct syntax is for this situation.

jackivanov:
We can escape it and keep the commas. Thanks!

Issue #1673, opened 2020-01-02
Name constraint on CA certificate


netheril96:
**Is your feature request related to a problem? Please describe.**
Currently, if we keep the PKI, we take the risk that someone steals the CA private key and uses it to sign server certificates our client devices will trust. They would be able to sign any server certificate, including well known ones like google.com, microsoft.com, apple.com, etc.

**Describe the solution you'd like**
Setting [name constraint](https://tools.ietf.org/html/rfc5280#section-4.2.1.10) on CA certificate so that it can only be used to sign our server certificate. Stealing CA private key will then result in no more damage than stealing the server certificate key.

**Describe alternatives you've considered**
We can choose to discard CA key, but then we cannot add more users.

**Additional context**

jackivanov:
Might be related #75

jackivanov:
Merged

Issue #75, opened 2016-08-21
Technically constrain root CA


ghost:
In the current form (from what I understand anyway) is that the root CA could be used to MITM HTTPS and other TLS connections that use the system certificate store (e.g. Google Chrome, iOS Safari) transparently if the box were to get popped (or your friend isn't one).

If the root were technically constrained there'd be less of a possibility to do damage to stacks that support the constraints if the private key were obtained.

Here's a selection of the constraints that could possibly be used but I'm sure I've missed or used some incorrectly:
- Basic Constraints, pathlen:0 - This prevents a subordinate/intermediate CA from being created under the root CA
- Name Constraints: This could be used to only allow issuance to the IP address of the server (permit: `IP:<server>`) This would probably affect client certificate issuance too but a randomly generated reserved "DNS name" might work for the permit side such as `DNS:<uuid4>` allowing `<client name>.<uuid4>` to be valid
-  Extended Key Usage: serverAuth, clientAuth, OCSPSigning(?)

leonelwilliams:
On macOS it looks like you can manually mark the root CA as not trusted and still have it work for the VPN connection (deleting the CA stops the connection from proceeding as expected)

In `Keychain Access.app` I opened the root and set `Trust -> When using this certificate:` to `Never Trust`. Unfortunately, iOS and presumably other platforms don't expose this setting

leonelwilliams:
Looks like macOS and iOS don't implement nameConstraints 😠

OpenSSL 1.0+ and Firefox seem to. Don't have anything else to test with

dguido:
$250 bounty! Submit a pull request and email dan@trailofbits.com to claim it. Partial solutions may be rewarded.

FiloSottile:
Wouldn't it be a simple, bare minimum mitigation to password-protect the CA private key? Or even better offer a mode in which it never touches the disk and then is destroyed?

The former doesn't help if the attacker waits for a users update, and neither help if a third party sets up a server for you, but this is a pretty big exposure increase. For instance, me myself I'd rather use a public wifi w/o VPN than allow a server to MitM my TLS connections.

FiloSottile:
The more I think about it the more I believe this warrants a pretty visible disclaimer. The threat model of who routes my traffic is directly informed by my trust in TLS verification.

dakami:
There's some deprecated EKU's around IPSec.

===
The following purposes have been included in a predecessor draft of RFC 3280 and therefore continue to be registrated by this implementation:
[...]

PR #1675, opened 2020-01-07
Ca certificate name constraints

jackivanov:
## Description
- changes the basic constraints to critical with `pathlen:0`:
```
X509v3 Basic Constraints: critical
    CA:TRUE, pathlen:0
```

- Generates a random id for every deployment and adds the nameConstraints extension to the CA certificate:
```
X509v3 Name Constraints: critical
    Permitted:
      IP:34.254.198.99/255.255.255.255
      DNS:585d3e61-7be1-5170-b417-e4e5d5b4fa57.algo
      email:585d3e61-7be1-5170-b417-e4e5d5b4fa57.algo   
```

## Motivation and Context
The discussion initially started in #75 and rose up again in #1673

## Types of changes

  - emails in usernames are not allowed anymore

## Checklist:

### DIFF
--- a/config.cfg
+++ b/config.cfg
@@ -4,6 +4,7 @@
 # Every device must have a unique username.
 # You can generate up to 250 users at one time.
 # Usernames with leading 0's or containing only numbers should be escaped in double quotes, e.g. "000dan" or "123".
+# Emails are not allowed
 users:
   - phone
   - laptop
--- a/roles/strongswan/defaults/main.yml
+++ b/roles/strongswan/defaults/main.yml
@@ -11,9 +11,9 @@ algo_dns_adblocking: false
 ipv6_support: false
 dns_encryption: true
 domain: false
-openssl_user_domain: algo.vpn
+openssl_constraint_random_id: "{{ IP_subject_alt_name | to_uuid }}"
 subjectAltName_IP: "{{ 'DNS:' if IP_subject_alt_name|regex_search('[a-z]') else 'IP:' }}{{ IP_subject_alt_name }}"
-subjectAltName_USER: "email:{{ item }}@{{ openssl_user_domain }}"
+subjectAltName_USER: "email:{{ item }}@{{ openssl_constraint_random_id }}"
 openssl_bin: openssl
 strongswan_enabled_plugins:
   - aes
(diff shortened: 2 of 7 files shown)


reaperhulk:
Can you tell me a bit more about the purpose of the randomized DNS/email constraints you've implemented here? Is it just so that a constraint for NC/email exists?

Also, even when IPv6 is not active a NC should be imposed for it -- you can do `IP:0:0:0:0:0:0:0:0/0:0:0:0:0:0:0:0` in excludedSubtrees to prohibit all IPv6.

jackivanov:
@reaperhulk I didn't want to stick to a static domain like `algo.local` or something similar, so randomizing is just a placeholder to enable the name constraints explicitly.  Thanks for pointing me to IPv6, I've fixed that, and also add an exclusion for all IPv4 range if the SAN is set to a domain.

reaperhulk:
Ah, okay. So you're using a permitted in all scenarios so that a NC is applied because `excludedSubtrees` works a bit weirdly for `dNSName` and `rfc822Name`. The CA/Browser Forum Baseline Requirements do document the correct `excludedSubtrees` model for `dNSName` to be an empty value, but my best guess for `rfc822Name` is `.` and it's safer in both cases to just have a positive constraint.

One last question: in the latest commit a positive name constraint is set on `dNSName` if and only if the `subjectAltName` type is `IP`:

```
  {%- if subjectAltName_type == 'IP' -%}
   ,permitted;DNS:{{ openssl_constraint_random_id }}
   {%- else -%}
   ,excluded;IP:0.0.0.0/0.0.0.0
   {%- endif -%}
```

Based on our conversation here I expected that a `dNSName` constraint would be set at all times. Why is it not?

jackivanov:
If `subjectAltName_type` is already set to DNS (a user configured the endpoint as a domain name) the `dNSName` will be set to this domain only. I'm not sure if we need to add a constraint for our own random domain in this case.

Here is what you'll see if the endpoint will be set to a domain:
```
X509v3 Name Constraints:
    Permitted:
      DNS:algo.com
      email:a806e892-4883-505a-8be7-a47a7f4015bd.algo
    Excluded:
      IP:0.0.0.0/0.0.0.0
      IP:0:0:0:0:0:0:0:0/0:0:0:0:0:0:0:0
```

reaperhulk:
Ah, okay, I understand now. This looks good at this point!

PR #1774, opened 2020-04-18
use ca_password from variable(--extra-vars) -  non-interactive installation using ansible playbook

saravanan30erd:
use ca_password from variable in non-interactive installation using ansible playbook.

## Description
https://github.com/trailofbits/algo/blob/master/docs/deploy-from-ansible.md#local-installation

I tried the algo installation non-interactively using ansible-playbook command directly as per the above documentation.

I noticed even if we set `ca_password` variable in ansible `--extra-vars` it still using the auto generated password for CA password.

```ansible-playbook main.yml -e "provider=local server=localhost endpoint=<public IP of VPN server> ca_password=xxxxxx"```

## Motivation and Context
It will use `ca_password` from variables if this variable has passed from `--extra-vars` or variable config file and only use auto-generated password if there is no `ca_password` variable configured.

It will enable users to provide `ca_password` when using [non-interactive installation](https://github.com/trailofbits/algo/blob/master/docs/deploy-from-ansible.md#local-installation) method.

## How Has This Been Tested?
- added test script to test this fix.

## Types of changes

## Checklist:

### DIFF
--- a/roles/common/tasks/facts.yml
+++ b/roles/common/tasks/facts.yml
@@ -6,7 +6,7 @@
 
 - name: Set facts
   set_fact:
-    CA_password: "{{ lookup('password', '/dev/null length=16 chars=ascii_letters,digits,_,@') }}"
+    CA_password: "{{ ca_password|default(lookup('password', '/dev/null length=16 chars=ascii_letters,digits,_,@')) }}"
     IP_subject_alt_name: "{{ IP_subject_alt_name }}"
 
 - name: Set IPv6 support as a fact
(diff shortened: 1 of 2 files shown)


jackivanov:
Actually an extra vars always win precedence, so you can just pass `CA_password`

saravanan30erd:
@jackivanov `CA_password` didn't work. `CA_password` is actually created by set_fact in `roles/common/tasks/facts.yml` using random auto-generation password.

Also `ca_password` properly works when adding new users by using `users,yml`

PR #1797, opened 2020-05-18
Refactor cloud-init/base.sh

djds:
## Description

* Pass shellcheck
* Use variable for username
* Fix issues with umask and sudo
* Simplify until loops

## Motivation and Context

A famous person once said "explicit is better than implicit." Because of
potential variations in cloud environments and options in `/etc/sudoers{,d}`,
such as `Defaults umask_override`, it was possible for `/home/algo/.ssh` and
`/home/algo/.ssh/authorized_keys` to be created with the wrong permissions.
This happened to me on VULTR, which led me to review the `cloud-init/bash.sh`
script in an effort to fix the issues that I encountered.

## How Has This Been Tested?

I tested my changes by deploying to VULTR from a virtual environment with the following commands:

```sh
python3 -m venv .venv
. .venv/bin/activate
pip install -U -r requirements.txt
./algo
```

## Types of changes

## Checklist:

### DIFF
--- a/files/cloud-init/base.sh
+++ b/files/cloud-init/base.sh
@@ -1,25 +1,41 @@
 #!/bin/bash
-set -eux
+set -euxo pipefail
 
-which sudo || until \
-  apt-get update -y && \
-  apt-get install sudo -yf --install-suggests; do
-  sleep 3
+readonly user='algo'
+
+export DEBIAN_FRONTEND='noninteractive'
+
+until which sudo; do
+    apt-get update -qq
+    apt-get install -qqf --install-suggests sudo
+    sleep 3
 done
 
-getent passwd algo || useradd -m -d /home/algo -s /bin/bash -G adm -p '!' algo
+getent passwd "${user}" \
+    || useradd -m -d "/home/${user}" -s /bin/bash -G adm -p '!' "${user}"
 
-(umask 337 && echo "algo ALL=(ALL) NOPASSWD:ALL" >/etc/sudoers.d/10-algo-user)
+(
+    umask 0337 \
+        && printf '%s\n' "${user} ALL=(ALL) NOPASSWD:ALL" \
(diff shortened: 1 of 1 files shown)


files/cloud-init/base.sh
```diff
@@ -1,25 +1,41 @@
...
 done
 
-getent passwd algo || useradd -m -d /home/algo -s /bin/bash -G adm -p '!' algo
+getent passwd "${user}" \
+    || useradd -m -d "/home/${user}" -s /bin/bash -G adm -p '!' "${user}"
+
+(
+    umask 0337 \
+        && printf '%s\n' "${user} ALL=(ALL) NOPASSWD:ALL" \
+        >"/etc/sudoers.d/10-${user}-user"
```
davidemyers, line 20: This is a *very* minor point, but I think the `sudoers.d` file name should contain the literal 'algo' rather than the variable, since that is more analogous to the `90-cloud-init-users` file created by `cloud-init`.

djds:
Is there anything else that I need to address before this can be merged?

davidemyers:
@caccavale I think this breaks Lightsail and should be reverted.

I tested it against Vultr but didn't think to try Lightsail.

davidemyers:
I think I found it. On Lightsail the script is apparently run by `dash` and not `bash` so `set -o pipefail` doesn't work. I'll do a little more testing and submit a PR.

jackivanov:
Well, we should have check all the providers before merging, not only lightsail might be broken now

davidemyers:
OK, I will not submit a new PR.

jackivanov:
@caccavale Please rollback it. I'll soon get more available time to test that

jackivanov:
Reverted. I'll soon get more time available to test it properly., have been quite busy last weeks

PR #1777, opened 2020-04-19
Ansible 2.8.8 -> 2.9.7 to fix python 3.8.2 errors

pawl:
## Description
Upgrades Ansible 2.8.8 -> ~~2.9.2~~ 2.9.7 in requirements.txt to fix issues with running Ansible on Python 3.8.2.

## Motivation and Context
I was having the issue described in: https://github.com/trailofbits/algo/issues/1622

```
ERROR! Unexpected Exception, this is probably a bug: cannot pickle '_io.TextIOWrapper' object
```

When attempting to run ./algo with Python 3.8.2 on MacOS 10.14.

## How Has This Been Tested?
I was able to run ./algo and complete an install successfully with the vultr playbook after this change.

## Types of changes
- Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [x] My code follows the code style of this project.
- [x] All new and existing tests passed.

PR #1838, opened 2020-07-08
Update macOS doc

davidemyers:
## Description

Update the macOS doc because Homebrew changed their Python package naming, and Algo is still using Ansible 2.8 which doesn't work with Python 3.8.

## How Has This Been Tested?

Works for me.

## Types of changes

## Checklist:

### DIFF
--- a/docs/deploy-from-macos.md
+++ b/docs/deploy-from-macos.md
@@ -32,13 +32,14 @@ The install command below takes care of initializing the CA certificate store.
 
 #### Installation
 ```
-brew install python3
+brew install python@3.7
+brew link --force python@3.7
 ```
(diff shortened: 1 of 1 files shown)


caccavale:
We should go the way of #1777 rather than recommending users downgrade their python versions.
Could you take a look at that PR?

davidemyers:
I agree that using a newer version of Ansible is the best solution, but since I wasn't sure when that would happen I wanted to update this doc to prevent Homebrew users from running into problems in the mean time.

#1777 is fine (though I believe Ansible is now up to 2.9.11), the issue is testing all of the supported cloud providers. I don't have accounts with all of them.

PR #1855, opened 2020-08-04
Discontinue use of the WireGuard PPA

davidemyers:
## Description

Discontinue use of the [WireGuard PPA](https://launchpad.net/~wireguard/+archive/ubuntu/wireguard).

## Motivation and Context

WireGuard is now included in the Ubuntu kernel package for all supported Ubuntu versions since 16.04 LTS. Therefore [the WireGuard PPA is being removed](https://lists.zx2c4.com/pipermail/wireguard/2020-August/005737.html).

## How Has This Been Tested?

Deployed to DigitalOcean and Vultr (where Algo uses 20.04) and Lightsail (where Algo still must use 18.04). Someone should test OpenStack (which still uses 18.04).

## Types of changes

- Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] All new and existing tests passed.

PR #1857, opened 2020-08-06
WireGuard is included in Ubuntu by default now

zx2c4:
WireGuard is now in Ubuntu 20.04, 18.04, and 16.04 in their default
kernels, and all other Ubuntus are EOL'd by Canonical. [1] has details
on replacing the PPA with Canonical's builtin repos, which this commit
implements.

[1] https://lists.zx2c4.com/pipermail/wireguard/2020-August/005737.html

---

I haven't tested this, so please test before merging. Also, everything usually works better after an `apt update && apt upgrade && reboot` as it usually takes Canonical a while to make Ubuntu work after the initial packages they release onto the ISO.

### DIFF
--- a/docs/client-linux-wireguard.md
+++ b/docs/client-linux-wireguard.md
@@ -5,11 +5,7 @@
 To connect to your AlgoVPN using [WireGuard](https://www.wireguard.com) from Ubuntu, first install WireGuard:
 
 ```shell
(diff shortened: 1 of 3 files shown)


davidemyers:
Jason, thanks for your work on WireGuard! But this PR looks redundant to #1855.

zx2c4:
Oh, excellent, thanks!

PR #1927, opened 2020-12-12
Explicitly set SSH permissions in base.sh

### LINKED ISSUE #1925: SSH To Vultr Servers Don't Work
I've repeated this issue on more than one server on Vultr. It works on Hetzner for example, but using the Shell access command: ssh -F config/SERVER_IP/ssh_config server_name the SSH to vultr fails with error: algo@192.248.177.217: Permission denied (publickey).

Appreciate if this could be looked at as currently don't have access to the server. Thanks.

davidemyers:
## Description
Apparently `umask` does not propagate through to `sudo` commands. This change explicitly sets the permissions on the `.ssh` directory and `authorized_keys` file for the user `algo`.

Also not all cloud providers obey the `#!` in a script like this so we must assume the script will be run by `dash` rather than `bash`.

Closes #1925.

## How Has This Been Tested?
Tested on Vultr, Lightsail, and Linode.

## Types of changes

## Checklist:

### DIFF
--- a/files/cloud-init/base.sh
+++ b/files/cloud-init/base.sh
@@ -1,6 +1,7 @@
-#!/bin/bash
+#!/bin/sh
 set -eux
 
+# shellcheck disable=SC2230
 which sudo || until \
   apt-get update -y && \
   apt-get install sudo -yf --install-suggests; do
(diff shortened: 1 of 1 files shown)


davidemyers:
I don't know why the test failed but I had a `pip install` fail for me at one point today as well.

SonyViao (approved):

PR #14114, opened 2020-12-28
adding pipenv support

### LINKED ISSUE #1868: Pipenv usage for repo
No, but it has been mentioned in this issue: https://github.com/trailofbits/algo/issues/1395

**Describe the solution you'd like**
I was curious if the project would be opposed if I submitted a pull request for generating files for using pipenv (i.e. Pipfile & Pipfile.lock), because you can still generate `requirements.txt` from those file (this command specifically: `pipenv lock -r`) but it also allows you to install dev packages. You could put ansible-lint, and if there are other, in there and then run  `pipenv install -d`, and that will install all the normal dependencies and dev ones as well.

This would allow the project to contain all python dependencies, even for development, and could be used for a very portable and   and more deterministic development/CI environments. This isn't required as you can see from these issues: https://github.com/trailofbits/algo/pull/863 and https://github.com/trailofbits/algo/issues/1554, but I think it would help make it more approachable as well as pipenv will manage virtual envs for the user instead of having to manually have them create it i.e.: step 4 in: https://github.com/trailofbits/algo#deploy-the-algo-server

**Describe alternatives you've considered**
Not really, and just curious specifically about pipenv, because I am not well versed in other package management tools.

**Additional context**
great video, by the author ( @ken-reitz ) of pipenv, talking about it and it's different features: https://youtu.be/GBQAKldqgZs

elreydetoda:
Thank you for this contribution! While pipenv was a good solution when this PR was created in 2020, the Python packaging ecosystem has evolved significantly since then.

We're closing this PR as the community has largely moved beyond pipenv to more modern solutions. We'll be exploring simpler packaging options that minimize setup requirements for non-developer users.

We appreciate your effort in improving Algo's developer experience!

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,15 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+ansible-core = "==2.11.3"
+ansible = "==4.4.0"
+netaddr = "*"
+Jinja2 = "~=3.0.1"
+
+[dev-packages]
+
+[requires]
+python_version = "3"
--- a/Pipfile
+++ b/Pipfile
@@ -4,10 +4,10 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-ansible-core = "==2.11.3"
-ansible = "==4.4.0"
+ansible-core = "*"
+ansible = "*"
 netaddr = "*"
-Jinja2 = "~=3.0.1"
+Jinja2 = "*"
 
 [dev-packages]
 
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,17 +18,17 @@
     "default": {
         "ansible": {
             "hashes": [
-                "sha256:1345fa61c9ac2edcc6c16e0d3f872a73303a61b799a7f16236ca6279e0ae8e36"
+                "sha256:4f19820ba2762e02051e6fdd00bf270b191990daf399687a32b06a0aa410360c"
             ],
             "index": "pypi",
-            "version": "==4.8.0"
+            "version": "==4.4.0"
         },
         "ansible-core": {
             "hashes": [
-                "sha256:93d50283c7c5b476debf83dc089b3f679b939a8b9a7b5d628d28daafbb3d303a"
+                "sha256:0ced1b4f67197edb27b50934c95d4cb644c6d635d72c7f826c4425dfe4537674"
             ],
             "index": "pypi",
-            "version": "==2.11.6"
+            "version": "==2.11.3"
         },
         "cffi": {
             "hashes": [
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "1277a5279f6bb88863ea7a08d9bf546dca26dab2c405640d45fc5b5f958f9ec8"
+            "sha256": "0c3c627d43a80cad0a384a195031ecdf469cea355e4b4e878b0a5f089e925d25"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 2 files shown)


Fuckmylimpdick:
No

elreydetoda:
> No

Ok, @Fuckmylimpdick is there any reason in particular as to why you are saying no?

jauderho:
+1 to this.

We can generate a matching requirements.txt from the Pipefile for folks that want to continue using that but in general, I've found that pipenv has been better at working out the constraints & dependencies.

elreydetoda:
It's been a bit since I have installed the deps for that branch, but I should easily be able to update this PR if someone from the project wants to accept it :slightly_smiling_face:

jauderho:
I'd say to wait after #14272 is merged

elreydetoda:
tl;dr: this has the exact same dependencies that the requirements.txt has in master currently.

So, this Pipfile & Pipfile.lock should be pinning to the exact dependencies inside the current [requirements.txt](https://github.com/trailofbits/algo/blob/4bed66f19eca9636528985cac5f6a725bc2c8739/requirements.txt).

I added an [intermediary commit](https://github.com/trailofbits/algo/pull/14114/commits/aa61ffcc8668ec4c3c7d1cdeed5135ee4b497a5a) in there, so I could add `"*"` for all the dependencies.  This will make upgrades in the future easier, but in the last commit you can see I reverted the hashes from the initial commit and updated the `hash` line specifically so Pipenv doesn't try to update the lock file again.


```console
$ gd HEAD..5210bf4 | cat
diff --git a/Pipfile b/Pipfile
index 3926440..8002695 100644
--- a/Pipfile
+++ b/Pipfile
@@ -4,10 +4,10 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-ansible-core = "*"
-ansible = "*"
+ansible-core = "==2.11.3"
+ansible = "==4.4.0"
 netaddr = "*"
-Jinja2 = "*"
+Jinja2 = "~=3.0.1"
 
 [dev-packages]
 
diff --git a/Pipfile.lock b/Pipfile.lock
index d5b4e73..bed9528 100644
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "0c3c627d43a80cad0a384a195031ecdf469cea355e4b4e878b0a5f089e925d25"
+            "sha256": "1277a5279f6bb88863ea7a08d9bf546dca26dab2c405640d45fc5b5f958f9ec8"
         },
         "pipfile-spec": 6,
         "requires": {
```


Speaking of updates, the way you should be able to update dependencies with pipenv is `pipenv update` command.

I say *should*, because ( like @glennschler had an issue with [here](https://github.com/trailofbits/algo/pull/14272#issuecomment-958134347) ) sometimes ansible wants a fresh install.  So, in that can I have had to run `pipenv --rm && pipenv install --skip-lock && pipenv lock`.  This will essentially install all packages "fresh" and then pin the dependencies after the install.

elreydetoda:
@dguido, next time would you mind just making your own comment at the bottom of the GitHub issue instead of directly editing my original comment? That was it keeps the context for anyone else if they stumble upon the issue. I personally know I can see the old version through the [edit history](https://github.blog/changelog/2018-05-24-comment-edit-history/), but that's not obvious to everyone.

Also, yes I completely understand that. I think a more modern solution would probably be like [Astral's `uv`](https://docs.astral.sh/uv/).
Has your team heard or considered that?

I see that the project already has a [pyproject.toml](https://github.com/trailofbits/algo/blob/0ddd994ff489a39a80d91ee7194618f744a807c1/pyproject.toml), so it could just be an extension of that. I'd be happy to do a quick PoC if you'd like?

The nice thing about uv as well is that it's cross platform, as it is built on rust, so that makes it work under Windows or Linux.

PR #14173, opened 2021-03-15
add sudo requirement for local install

destan:
add sudo requirement for local install in the README

## Description

It fails with following error if ./algo is not run with sudo:

fatal: [localhost]: FAILED! => {"changed": false, "msg": "non-zero return code", "rc": 1, "stderr": "sudo: a password is required\n", "stderr_lines": ["sudo: a password is required"], "stdout": "", "stdout_lines": []}

This happens when the algo script is run on the target machine.

I assume this would still happen if algo script run from another machine but the ssh user has no sudo privilege or requires a password for sudo. However, I didn't test this case hence didn't add anything about that. 

## How Has This Been Tested?

Tested with Ubuntu 20.04 updated today (2021-03-15) and algo repository with commit id `70f9f91`.

## Types of changes

- Readme update

## Checklist:


dguido:
Thank you! Closing in favor of #14790

Issue #14234, opened 2021-06-27
dnscrypt-proxy fails to start - signature problem


iamvishnurajan:
**Describe the bug**

After creating an algo VPN instance on DigitalOcean, I was having frequent dns lookup problems. After logging into the server, I see that dnscrypt-proxy failed to start. Logging shows there is some kind of a signature issue.

algo@AlgoVPN-NYC3-20210627:~$ systemctl --failed
  UNIT                   LOAD   ACTIVE SUB    DESCRIPTION          
● dnscrypt-proxy.service loaded failed failed DNSCrypt client proxy

LOAD   = Reflects whether the unit definition was properly loaded.
ACTIVE = The high-level unit activation state, i.e. generalization of SUB.
SUB    = The low-level unit activation state, values depend on unit type.

algo@AlgoVPN-NYC3-20210627:~$ journalctl -u dnscrypt-proxy
-- Logs begin at Sun 2021-06-27 05:54:40 UTC, end at Sun 2021-06-27 06:04:07 UTC. --
Jun 27 05:57:31 AlgoVPN-NYC3-20210627 systemd[1]: Started DNSCrypt client proxy.
Jun 27 05:57:31 AlgoVPN-NYC3-20210627 dnscrypt-proxy[2981]: [2021-06-27 05:57:31] [NOTICE] dnscrypt-proxy 2.0.31
Jun 27 05:57:31 AlgoVPN-NYC3-20210627 dnscrypt-proxy[2981]: [2021-06-27 05:57:31] [NOTICE] Network connectivity detected
Jun 27 05:57:32 AlgoVPN-NYC3-20210627 dnscrypt-proxy[2981]: [2021-06-27 05:57:32] [NOTICE] Source [/var/cache/dnscrypt-proxy/public-resolvers.md] loaded
Jun 27 05:57:32 AlgoVPN-NYC3-20210627 dnscrypt-proxy[2981]: [2021-06-27 05:57:32] [NOTICE] Firefox workaround initialized
Jun 27 05:57:32 AlgoVPN-NYC3-20210627 dnscrypt-proxy[2981]: [2021-06-27 05:57:32] [WARNING] Systemd sockets are untested and unsupported - use at your own r>
Jun 27 05:57:32 AlgoVPN-NYC3-20210627 dnscrypt-proxy[2981]: [2021-06-27 05:57:32] [NOTICE] Wiring systemd TCP socket #0, dnscrypt-proxy.socket, 127.0.2.1:53
[...]

PR #1334, opened 2019-02-20
Refactoring

jackivanov:
## Description
Renames the vpn role to strongswan, and split up the variables to support 2 separate VPNs. Closes #1330 and closes #1162
Configures Ansible to use python3 on the server side. Closes #1024 
Removes unneeded playbooks, reorganises a lot of variables
Reorganises the `config` folder. Closes #1330
<details><summary>Here is how the config directory looks like now</summary>
<p>

```
configs/X.X.X.X/
|-- ipsec
|   |-- apple
|   |   |-- desktop.mobileconfig
|   |   |-- laptop.mobileconfig
|   |   `-- phone.mobileconfig
|   |-- manual
|   |   |-- cacert.pem
|   |   |-- desktop.p12
|   |   |-- desktop.ssh.pem
|   |   |-- ipsec_desktop.conf
|   |   |-- ipsec_desktop.secrets
|   |   |-- ipsec_laptop.conf
|   |   |-- ipsec_laptop.secrets
|   |   |-- ipsec_phone.conf
|   |   |-- ipsec_phone.secrets
|   |   |-- laptop.p12
|   |   |-- laptop.ssh.pem
|   |   |-- phone.p12
|   |   `-- phone.ssh.pem
|   `-- windows
|       |-- desktop.ps1
|       |-- laptop.ps1
|       `-- phone.ps1
|-- ssh-tunnel
|   |-- desktop.pem
|   |-- desktop.pub
|   |-- laptop.pem
|   |-- laptop.pub
|   |-- phone.pem
|   |-- phone.pub
|   `-- ssh_config
`-- wireguard
    |-- desktop.conf
    |-- desktop.png
    |-- laptop.conf
    |-- laptop.png
    |-- phone.conf
    `-- phone.png
```

finder

</p>
</details>

## Motivation and Context
This refactoring is focused to aim to the 1.0 release

## How Has This Been Tested?
Deployed to several cloud providers with various options enabled and disabled

## Types of changes

- [x] Refactoring

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] All new and existing tests passed.

PR #14235, opened 2021-06-29
Move the dnscrypt-proxy cache file

davidemyers:
## Description
Move the `dnscrypt-proxy` cache file of public resolvers from `/tmp` to `/var/cache/dnscrypt-proxy`.

## Motivation and Context
Several years ago in #1120 I moved the `dnscrypt-proxy` cache file of public resolvers to `/tmp`. The default had been to write it to `/etc/dnscrypt-proxy` but this was failing because `dnscrypt-proxy` does not run as `root` and also is blocked from writing to that directory by AppArmor.

Recently, as reported in #14234, there was a problem with the signing of the public resolver file which prevented `dnscrypt-proxy` from starting during Algo deployment. This could have also affected any existing Algo servers that were rebooted while the error was present since the cache file in `/tmp` would have been lost after a reboot.

According to the author of `dnscrypt-proxy` [here](https://github.com/DNSCrypt/dnscrypt-resolvers/issues/519#issuecomment-869169729), recent versions will ignore problems fetching the public resolver file and continue to use the cached version.

Therefore this change moves the cache to the Ubuntu package location of `/var/cache/dnscrypt-proxy` and adjusts the AppArmor rules. The PPA version of `dnscrypt-proxy`, which is still used on Ubuntu 18.04 deployments, uses `/var/cache/private/dnscrypt-proxy`, but creates `/var/cache/dnscrypt-proxy` as a symbolic link to that directory.

This change also updates the `dnscrypt-proxy` config to eliminate the warning message:
```
config option `refused_code_in_responses` is deprecated, use `blocked_query_response`
```

## How Has This Been Tested?
Deployed to Vultr with Ubuntu 18.04 and 20.04.

## Types of changes

## Checklist:

### DIFF
--- a/roles/dns/files/apparmor.profile.dnscrypt-proxy
+++ b/roles/dns/files/apparmor.profile.dnscrypt-proxy
@@ -14,7 +14,7 @@
 
   /etc/dnscrypt-proxy/** r,
   /usr/bin/dnscrypt-proxy mr,
-  /tmp/public-resolvers.md* rw,
+  /var/cache/{private/,}dnscrypt-proxy/** rw,
 
   /tmp/*.tmp w,
   owner /tmp/*.tmp r,
--- a/roles/dns/templates/dnscrypt-proxy.toml.j2
+++ b/roles/dns/templates/dnscrypt-proxy.toml.j2
@@ -118,11 +118,12 @@ timeout = 2500
 keepalive = 30
 
 
-## Use the REFUSED return code for blocked responses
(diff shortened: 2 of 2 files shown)


Bow0628:
@davidemyers # # #14235 #1334 
Duplicate of #

Bow0628 (approved):
@attilagyorffy #14235 @statik #
Duplicate of #
Duplicate of #

Issue #14286, opened 2021-09-03
Not able to view Twitter Profiles and Tweet Conversations/Threats on iOS


domdewom:
### **OS / Environment (where do you run Algo on)**
t2.micro Linux/UNIX Ubuntu

### **Cloud Provider (where do you deploy Algo to)**
AWS

### **Summary of the problem**
When connected to my AlgoVPN and using the Twitter app on iOS, loading user profiles, viewing conversations or threats as well as members of a conversation is not possible. Any other page or app seems to work without issues. I doubt it has to do with connection strength even though I am based in China... But any help would be appreciated.

### **Steps to reproduce the behavior**
**Loading user profile**
1. While connected to AlgoVPN, open Twitter on iOS.
2. Click on any name of the user to see their profile
3. I get a blank profile page consisting of nothing but the user's Twitter handle and a message saying "Something went wrong, try again.", see picture below

**Loading conversation/threat**
1. While connected to AlgoVPN, open Twitter on iOS.
2. Find a tweet with a conversation/replies or with a threat.
3. Click on tweet to open conversation/threat
4. I get a page only opneing the original tweet without anything more

On other VPNs (commercial or corporate) Twitter App on iOS works perfectly fine and without above issues.

![algovpn](https://user-images.githubusercontent.com/517299/132074755-a7aa0a84-dc29-4c60-a5ec-86607162de8f.png)

davidemyers:
I've run into this problem in the past when using Twitter in a browser and I think it's due to Twitter blocking connections from certain cloud provider networks. But for some reason I no longer have this problem even though I'm also using Amazon (in my case Lightsail in region `us-east-1b`).

I've never encountered this problem when using [Tweetbot](https://apps.apple.com/us/app/tweetbot-6-for-twitter/id1527500834) on either iOS or macOS.

samkelleher:
@domdewom What service provider did you use?

I use DigitalOcean and have had the same instance running for about 2 years now, and in the last week or so I've had a tonne of issues.

I don't think it's related to certain blocking of cloud provider networks as it'll work randomly, but rather some something else.

The reason I say that is that the failure of twitter images (and just CDN served assets in general) is that they will eventually work if for example you restart the browser, or even refreshing the page a few times.

The network traffic is failing to do DNS lookups, rather than connecting and being blocked.

```ERR_NAME_NOT_RESOLVED``` is the error I always see in the traffic for domains like `pbs.twimg.com`

[...]

PR #14338, opened 2021-10-31
Change DNS from Cloudflare to Google

davidemyers:
## Description
Change the default encrypted DNS provider from Cloudflare to Google.

## Motivation and Context
Cloudflare's DNS-over-HTTPS service has become unreliable and is causing issues for Algo users. See #14286 and #14308.

Personally I prefer Quad9 but they have a bad habit of changing the symbolic names of their servers and breaking things. See #14269 and #14167.

## How Has This Been Tested?
Deployed to Vultr.

## Types of changes

## Checklist:

### DIFF
--- a/config.cfg
+++ b/config.cfg
@@ -88,12 +88,14 @@ dnscrypt_servers:
   ipv4:
     - google
 #   - cloudflare
+#   - quad9-doh-ip4-port443-filter-pri
 #   - <YourCustomServer>  # E.g., if using NextDNS, this will be something like NextDNS-abc123.
                           # You must also fill in custom_server_stamps below. You may specify
                           # multiple custom servers.
   ipv6:
-    - google-ipv6
+#   - google-ipv6
 #   - cloudflare-ipv6
+#   - quad9-doh-ip6-port443-filter-pri
 
 custom_server_stamps:
 # YourCustomServer: 'sdns://...'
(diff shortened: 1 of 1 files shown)


davidemyers:
I don't know if it's my recent changes to the tests or this PR, but I've noticed random test failures. Sometimes they pass and sometimes they don't.

davidemyers:
It looks like the failures occur when `dnscrypt-proxy` chooses `google-ipv6` over `google`, so I commented out `google-ipv6` and also added commented out entries for Quad9.

So Cloudflare is broken, as is Google's IPv6 server, and Quad9 likes to change their server names. Is there a better default for Algo users?

jackivanov:
Well, we can drop upstream ipv6 I believe, we've encountered problems with dnscrypt-proxy and ipv6 before too

mikedizon:
Twitter is still blocking certain requests when using google (with google-ipv6 as well) in their native apps. Web seems to be fine.  How can I view the request logs on the instance as they come in? I'm open to trying a few different providers listed on https://github.com/DNSCrypt/dnscrypt-resolvers/blob/master/v2/public-resolvers.md until I find one that is not being blocked.

davidemyers:
This PR has been open for long enough without acceptance or feedback so I'm treating it as rejected and closing it.

PR #14398, opened 2022-01-05
fix digitalocean by updating ipv4 address json path

incertia:
the current api format is documented at https://docs.digitalocean.com/reference/api/api-reference/#operation/get_droplet

if it changes again one can just update the path with some profound observations


incertia:
wait we fixed this already

PR #14411, opened 2022-01-29
Update Cloud Shell instructions

davidemyers:
## Description

Now that Algo requires Python 3.8 it no longer works from Google Cloud Shell. This PR notes this in the documentation.

## Checklist:


bn-l:
Note: I was having this error with python3.6 and python3.7. I install python 3.8 but then python3 was still aliased to 3.6.9 (even after running apt-remove on it). This stackexchange answer helped me alias python3 to python3.8: https://unix.stackexchange.com/a/520136 (which made the copy and pasted commands work).

PR #14536, opened 2022-10-09
Use cheapest EC2 instance as default

ali92hm:
## Description

Add documentation for using the AWS graviton instances that are generally cheaper.
More info on the motivation and context below, I'm also open to suggestions.

## Motivation and Context

For existing AWS customers or the accounts that have used their first year "free credits", the cheapest EC2 option is the graviton (arm64) EC2 class with the smallest size `t4g.nano`. I used this instance size with spot instances for a couple of months for as low as $2/month (everything included)

Also added a section about the AWS promotion on `t4g.small` instances.

## How Has This Been Tested?

Deployed an instance to my aws account.

## Types of changes

- [x] Documentation change 

These changes do not modify any existing instances, but the default behavior for the new EC2 instances launched will be different as the default size and architecture have changed. This could be viewed as a breaking change.

## Checklist:

- [ ] My code follows the code style of this project.

### DIFF
--- a/docs/cloud-amazon-ec2.md
+++ b/docs/cloud-amazon-ec2.md
@@ -6,13 +6,17 @@ Creating an Amazon AWS account requires giving Amazon a phone number that can re
 
 ### Select an EC2 plan
 
-The cheapest EC2 plan you can choose is the "Free Plan" a.k.a. the "AWS Free Tier." It is only available to new AWS customers, it has limits on usage, and it converts to standard pricing after 12 months (the "introductory period"). After you exceed the usage limits, after the 12 month period, or if you are an existing AWS customer, then you will pay standard pay-as-you-go service prices.
+The default EC2 plan (`t4g.nano` arm64) is the cheapest non "Free Plan" option. If you have a new AWS account or are eligible for the "AWS Free Tier" plan and would like to use that plan follow the instructions below.
+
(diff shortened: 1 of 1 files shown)


glennschler:
I agree it is a good idea, though some may want to fully understand they are using wireguard on ARM. I do not recall the complete history, though feel it has been available for only a few years vs Intel. At least with Ubuntu.

Secondly, AWS is on the 3rd year free trail of one t4g.small for everyone, regardless of if your 12-month free trial is over. This is what I have been using as one full time reserved instance, and then spot instances for others. https://aws.amazon.com/ec2/faqs/#t4g-instances

ali92hm:
Ahh that's good to know, sad that it will only go to the end of this year, but I can make the change to `t4g.small` instead. 

Regarding the arm vs x86, did you just wanted me to add a note on the read me?

glennschler:
The t4g.small free promotion has been extended the past two Decembers, so it is possible AWS does that again this December for another free year.

When the Algo project was created, the default image size may have been chosen not only based on price, but also performance. I feel it is best to not change the default, but yes a document comment can explain there are other choices to get the best price, especially highlighting the new ARM64 choices which did not exist in the past.

glennschler:
I do not believe the default EC2 should be changed to t4g.nano. A document comment is a better idea to explain there are other choices to get the best price, especially highlighting the new t4g ARM64 choices which did not exist in the past.

ali92hm:
Apologies, will make the change. It's been a little busy lately

glennschler:
Well I was not specific enough in my previous comments, plus It took some time to think about the issue. Still it's only my opinion which might not be the same as others.

ali92hm:
@glennschler made the changes you suggested and updated the PR description

glennschler:
LGTM

ali92hm:
I think I do need your approval :)

ali92hm:
Resolved a conflict

jackivanov (approved):

PR #14698, opened 2024-02-27
Fix hetzner module

jackivanov:
## Types of changes

## Checklist:

### DIFF
--- a/roles/cloud-hetzner/tasks/main.yml
+++ b/roles/cloud-hetzner/tasks/main.yml
@@ -1,12 +1,9 @@
 ---
-- name: Build python virtual environment
-  import_tasks: venv.yml
-
 - name: Include prompts
   import_tasks: prompts.yml
 
 - name: Create an ssh key
-  hcloud_ssh_key:
+  hetzner.hcloud.ssh_key:
     name: algo-{{ 999999 | random(seed=lookup('file', SSH_keys.public)) }}
     public_key: "{{ lookup('file', SSH_keys.public) }}"
     state: present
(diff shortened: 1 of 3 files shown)


RJNawara (approved):

roles/cloud-hetzner/tasks/main.yml
```diff
@@ -1,20 +1,17 @@
 ---
-- name: Build python virtual environment
-  import_tasks: venv.yml
-
```
ret2libc, line 4: Why was this removed together with the `venv.yml`? While testing this I could not proceed with the algo configuration because of various missing python libs, however the error went away after re-introducing these lines.

ret2libc: One of the error was:
```
TASK [cloud-hetzner : Get regions] *********************************************************************************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to import the required Python library (python-dateutil) on .. Python projects/algo/.env/bin/python3. Please read the module documentation and install it in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter"}
```

ret2libc:
Apart from the venv issue, the fix seems to work!

roles/cloud-hetzner/tasks/main.yml
jackivanov: Well yeah, apparently we still need the Python libraries. I mistakenly assumed that they come with the collection.

ret2libc (approved):

PR #14725, opened 2024-06-05
Update config.cfg: t2.micro to t3.micro

vitaly-zdanevich:
Because

1) t2.micro is not exists on every region, see "Some regions like the Middle East (Bahrain) region and the EU (Stockholm) region do not offer t2.micro instances" https://aws.amazon.com/free/free-tier-faqs/

2) t3.micro is faster https://www.cloudzero.com/advisor/t2-vs-t3/

Works for me
image

### DIFF
--- a/config.cfg
+++ b/config.cfg
@@ -186,7 +186,7 @@ cloud_providers:
     # Set use_existing_eip to "true" if you want to use a pre-allocated Elastic IP
     # Additional prompt will be raised to determine which IP to use
     use_existing_eip: false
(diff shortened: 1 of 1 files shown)


vitaly-zdanevich:
Even more - t2.micro is not exists in the UI for me 
image

jackivanov:
t3.micro is only free tier in regions where t2.micro isn't available. As long as this is true, we don't want to change the default. Anyone can update the config locally to use t3 if they want.

Issue #14726, opened 2024-06-11
Update-users failure (`Could not find or access 'configs/...`)


RochMoreau:
**Describe the bug**

When trying to update users by calling `./algo update-users`, the CLI throws an error.

I am on MacOS M1, the remote server is a custom Ubuntu 22.04

**To Reproduce**

1. Have an existing VPN server (configured with Algo, ensure a directory is correctly set in config/)
2. call `./algo update-users`
3. Select option `1`, corresponding to the server to be updated

**Expected behavior**

The process continues and updates users on the selected server

**Additional context**

Observing the behavior, I see that the options for the server-selection step includes the IP repeated 2 times. This string is then used as the server identifier when fetching the server config. The bug comes from this.
In short, let's say the server IP is `x.x.x.x`. There is a directory `config/x.x.x.x`. The CLI gives the following choice:
```
1. x.x.x.xx.x.x.x
```
(notice how the IP is repeated twice). When the user selects the option 1, instead of just using the ip, the CLI uses the full string, i.e. the IP repeated twice. The CLI now looks for the server config at `config/x.x.x.xx.x.x.x/` instead of `config/x.x.x.x`

The line causing this is line 30 from `users.yml`: `'{{ config.IP_subject_alt_name }}'`
I tracked down the introduction of this behavior in the PR #14718 from may 5 2024, so about 1 month old.

**Full log**

```
[Server address prompt]
Select the server to update user list below:
    1. x.x.x.xx.x.x.x

:

TASK [Server address prompt] ****************************************************************************************************************************************************
ok: [localhost]

TASK [Set facts based on the input] *********************************************************************************************************************************************
ok: [localhost]

TASK [Import host specific variables] *******************************************************************************************************************************************
[...]

PR #14718, opened 2024-05-06
Use legacy OpenSSL Format for Apple Devices

vvorlov:
## Description
Uses legacy key format to generate `.p12` and `.mobileprovision` files. 

## Motivation and Context
`openssl@3` uses a modern private key format unsupported on Apple devices because Apple has not updated its libraries. Importing generated `.p12` files into the modern macOS and iOS keychain leads to the wrong password error. If you try to export a private key from `.p12` and import it into Keychain separately, it will show the wrong format error. 

Closes #14558

Based on https://github.com/trailofbits/algo/pull/14622 with applied comments.

## How Has This Been Tested?
1. Amazon Lighstail
2. Cloned repo and installed via `METHOD=local ENDPOINT=<ip_adderess> USERS=hello,world ./install.sh`. I had to modify the install script to point to the fork on the `getAlgo` [step](https://github.com/trailofbits/algo/blob/6ce6f5c81e8e5a27ae6f4357512ee2e3362605aa/install.sh#L30)
3. Downloaded `.mobileprovision` files
4. Installed on macOS 14.4.1 (23E224)
5. Connected to the server.
6. Installed on iOS 17.4.
5. Connected to the server.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [x] I have read the **CONTRIBUTING** document.
- [x] My code follows the code style of this project.
- [] My change requires a change to the documentation.
- [] I have updated the documentation accordingly.
- [] I have added tests to cover my changes.
- [] All new and existing tests passed.

PR #14727, opened 2024-06-11
fix: fix server selection in update-user

RochMoreau:
Fixes server selection in update-user while preserving a nice display of server along with its alt_name in the list

## Description

This commit fixes a bug preventing correct selection of server when trying to update users. It improves the prompt's clarity by providing both server name and IP_subject_alt_name. It also ensures server selection from the list uses actual server names instead of list descriptions strings that caused the initial bug.

The previous code was using directly the strings generated from the "Build list of installed servers" as server identifier. 
A previous commit (#14718) introduced some display-related text in the string provoking errors at the step "Server address prompt" since the provided strings no longer represent server identity.

2 possible choices:
1. Keep the initial system of passing strings from "Build list of installed servers" to "Server address prompt"
  1. Benefits: It works, it was the initial behavior before #14718 
  2. Drawback 1: Cannot display some helpful data in the prompt like the IP_subject_alt_name of the server
  3. Drawback 2: Prone to the same error in the future if somebody adds new display-related data in the strings
2. Make "Build list of installed servers" return an array of object and "Server address prompt" pick only relevant properties for display, and relevant properties for server selection
  1. Benefit 1: Future-proof, the code is clear in the intent and the risk of re-introducing a breaking change is low
  2. Benefit 2: Keeps a nice display of the servers for selection, using `IP (name)` 
  3. Drawback: The code is a bit more complex since it now handles objects instead of plain strings, but this could count as a benefit too because now the manipulated data is explicitly named with object keys

This PR went with the option 2

## Motivation and Context
Solves issue #14726 

## How Has This Been Tested?
Manually tested using the same reproduction steps provided in issue #14726 

## Types of changes

## Checklist:

### DIFF
--- a/users.yml
+++ b/users.yml
@@ -25,19 +25,17 @@
           set_fact:
             server_list: >-
               [{% for i in _configs_list.files %}
-                {% set config  = lookup('file', i.path)|from_yaml %}
-                '{{ config.server }}'
-                '{{ config.IP_subject_alt_name }}'
-                {{ ',' if not loop.last else '' }}
+                {% set config = lookup('file', i.path) | from_yaml %}
+                {{ {'server': config.server, 'IP_subject_alt_name': config.IP_subject_alt_name} }}
               {% endfor %}]
 
         - name: Server address prompt
           pause:
             prompt: |
-              Select the server to update user list below:
+              Select the server to update user list below: 
                 {% for r in server_list %}
-                {{ loop.index }}. {{ r }}
-                {% endfor %}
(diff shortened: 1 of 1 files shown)


ddxor:
I manually applied this fix to my instance and it fixed this issue for me. Thanks @RochMoreau!

george2781 (approved):
This fixes my install when I apply it so looks good to me

jackivanov (approved):

PR #14747, opened 2024-09-21
Feature: Add scripts for easier user management (add/remove users)

velrino:
## Description
I've added two bash scripts to simplify the user management process in AlgoVPN.

The first script allows for automatically adding new users to the configuration file.
The second script makes it easy to remove existing users from the configuration file.
These scripts help avoid manual edits to the configuration file, making user management more efficient and reducing the risk of errors

## Motivation and Context
My motivation for adding these scripts was to streamline the process of managing users in AlgoVPN. By automating the addition and removal of users in the configuration file, the process becomes faster and less error-prone, improving efficiency in managing VPN users

## How Has This Been Tested?
Both scripts have been tested locally by adding and removing several users from the configuration file. The scripts were executed in a variety of scenarios to ensure that users are added correctly and removed without affecting other parts of the configuration file.

When run `./add-user.sh example`

image

When run `./remove-user.sh example`

image

## Types of changes

## Checklist:

### DIFF
--- a/add-user.sh
+++ b/add-user.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# run
+# chmod +x ./add-user.sh 
+# ./add-user.sh example
+
+# Checks if the username was passed as an argument
+if [ -z "$1" ]; then
+  echo "Usage: $0 <username>"
+  exit 1
+fi
+
+USERNAME=$1
+
+# Define the path to the configuration file
+CONFIG_FILE="./config.cfg"
+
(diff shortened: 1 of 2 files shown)


jackivanov:
Updating a YAML file is already quite easy, so I don’t see why there’s a need for such a script. Furthermore, `yq` can handle everything efficiently.

```shell
yq -i 'del(.users[] | select(. == "phone"))' config.cfg
yq -i '.users += ["new-phone"]' config.cfg
```

PR #14786, opened 2025-08-03
Fix IPv6 address selection on BSD systems

### LINKED ISSUE #1843: Algo ipv6 support for *bsd
Ansible maintainers appear to not care at all that ipv6 default isn't useable, this bug report has been open for 4 years.

ansible_default_ipv6 fact uses link-local address #16977 - https://github.com/ansible/ansible/issues/16977

algo ipv6 address needs to either be configurable, or heuristic applied to ansible addresses.

The site link local address messes up openssl certificate creation because of the '%ifacename' addition, but what's really needed at that point is the well known global address anyway.

dguido:
## Summary

This PR fixes IPv6 support on BSD systems (FreeBSD, OpenBSD, NetBSD) by properly selecting global IPv6 addresses instead of link-local addresses for certificate generation.

## Problem

BSD systems return IPv6 addresses in the order they were added to the interface, not sorted by scope like Linux does. This causes `ansible_default_ipv6` to often contain link-local addresses (fe80::) with interface suffixes (%em0) instead of global addresses. This breaks certificate generation because:

1. Link-local addresses include a `%interface` suffix that's invalid in certificates
2. Link-local addresses aren't routable and shouldn't be used for VPN endpoints

The upstream Ansible issue [#16977](https://github.com/ansible/ansible/issues/16977) has been open since 2016 without resolution.

## Solution

This PR implements a workaround in Algo:

1. Added `bsd_ipv6_facts.yml` that filters IPv6 addresses to find the first global address
2. Integrated this fix into the FreeBSD setup flow
3. The fix properly handles both interface-specific and system-wide IPv6 address lists
4. Falls back gracefully when IPv6 isn't available

## Testing

- Added a test playbook `tests/test_bsd_ipv6.yml` that verifies the address selection logic
- The fix only runs on BSD systems when IPv6 is enabled
- No changes to non-BSD systems

Fixes #1843


jackivanov:
@claude rebase, fix conflicts in tests/unit/test_template_rendering.py and add a new line at the end of roles/common/tasks/bsd_ipv6_facts.yml

jackivanov:
@claude to resolve conflicts: Create a fresh branch from master, include fixes from this branch and create a new PR

PR #14837, opened 2025-09-15
Fix SSH configuration indentation in cloud-init template

dguido:
## Summary
- Fixed SSH daemon startup failure caused by incorrect indentation in the sshd_config file
- Changed `indent(width=6, first=True)` to `indent(width=6, first=False)` in cloud-init template
- Resolves SSH connection timeouts during DigitalOcean (and potentially other cloud provider) deployments

## Problem
The SSH configuration file (`/etc/ssh/sshd_config`) was being written with 6 spaces of indentation on every line due to the `first=True` parameter in the Jinja2 indent filter. This made the configuration file invalid, causing the SSH daemon to fail during startup and preventing connections on the configured port (4160).

## Solution
Removed the `first=True` parameter (equivalent to `first=False`) from the indent filter to ensure the SSH configuration is written with proper formatting that sshd can parse.

## Test Plan
- [ ] Deploy a new Algo server on DigitalOcean
- [ ] Verify SSH connection succeeds on port 4160
- [ ] Confirm the deployment completes successfully
- [ ] Check that `/etc/ssh/sshd_config` on the deployed server has correct formatting

This bug likely affects all cloud deployments that use cloud-init for initial configuration.

### DIFF
--- a/files/cloud-init/base.yml
+++ b/files/cloud-init/base.yml
@@ -35,7 +35,7 @@ users:
 write_files:
   - path: /etc/ssh/sshd_config
     content: |
-{{ lookup('template', 'files/cloud-init/sshd_config') | indent(width=6, first=True) }}
+{{ lookup('template', 'files/cloud-init/sshd_config') | indent(width=6, first=False) }}
 
 runcmd:
   - set -x


dguido:
## Analysis: This PR Should Not Be Merged

After thorough testing, I've determined that the proposed change (`first=True` → `first=False`) is **incorrect** and would actually **break** cloud-init parsing entirely.

### Why `first=True` is Correct

The cloud-init YAML uses a block scalar (`|`) for the sshd_config content:

```yaml
write_files:
  - path: /etc/ssh/sshd_config
    content: |
      Port 4160
      AllowGroups algo
```

With `first=True`, every line (including the first) gets 6-space indentation, which is **required** for valid YAML block scalar syntax.

### What Happens with `first=False`

```yaml
write_files:
  - path: /etc/ssh/sshd_config
    content: |
Port 4160        # ← No indentation breaks YAML parsing\!
      AllowGroups algo
```

This produces a YAML parse error:
```
could not find expected ':'
```

### Key Insight: Block Scalars Strip Indentation

YAML block scalars (`|`) automatically strip the leading indentation when parsing. So:
- **In the YAML file**: Content has 6-space indent (for valid YAML structure)
- **After cloud-init parses it**: The file written to `/etc/ssh/sshd_config` has **no** leading spaces

I verified this by rendering the full template - the sshd_config written to the server is clean:
```
Port 4160
AllowGroups algo
PermitRootLogin no
...
```

### The Real Issue

If SSH is failing on deployments, the root cause is **not** the indent filter. You mentioned verifying the issue persists on master - could you share:
1. The actual error from `/var/log/cloud-init-output.log`?
2. The contents of `/etc/ssh/sshd_config` on the failed server?
3. Output of `systemctl status sshd`?

This would help identify the actual cause of the SSH failures.

PR #14949, opened 2026-01-30
fix(aip): correct assertion for SNAT IPv4 address validation

genevera:
## Description
Ensure the SNAT IPv4 validation uses a boolean test so `assert` no longer fails on a string result.

## Motivation and Context
Fixes #14948. The assertion was using a filter that returns a string, which caused Ansible to error when evaluating the condition.

## How Has This Been Tested?
I ran it and it didn't crash.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [x] I have read the **CONTRIBUTING** document.
- [ ] My code passes all linters (`./scripts/lint.sh`)
- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] Dependencies use exact versions (e.g., `==1.2.3` not `>=1.2.0`).

PR #14950, opened 2026-01-30
Fix AIP boolean assertion for Ansible 12 jinja2_native mode

### LINKED ISSUE #14948: AIP functionality broken
**What happened?**
When provisioning a VM in DigitalOcean, the AIP task fails due to an assertion receiving a string `True` rather than a boolean

**Environment** (cloud provider, OS, WireGuard or IPsec)
DigitalOcean, Linux, WG

**Output**
```
Task failed: Conditional result (True) was derived from value of type 'str' at '/path/to/algo/roles/common/tasks/aip/main.yml:14:11'. Conditionals must have a boolean result
```

dguido:
## Summary

- Fix `assert.that` clause in AIP task that fails with Ansible 12+ due to `jinja2_native` mode
- The `ansible.utils.ipv4` filter returns the IP string (not a boolean), which causes "Conditional result was derived from value of type 'str'" error
- Add `| bool` filter to convert the truthy string to a proper boolean

Fixes #14948

## Test plan

- [x] `ansible-playbook main.yml --syntax-check` passes
- [x] `ansible-lint roles/common/tasks/aip/main.yml` passes (0 failures, 0 warnings)
- [x] `pytest tests/unit/ -q` passes (91 tests)

### DIFF
--- a/roles/common/tasks/aip/main.yml
+++ b/roles/common/tasks/aip/main.yml
@@ -11,5 +11,5 @@
 
 - name: Verify SNAT IPv4 found
   assert:
-    that: snat_aipv4 | ansible.utils.ipv4
+    that: (snat_aipv4 | ansible.utils.ipv4) | bool
     msg: The SNAT IPv4 address not found. Cannot proceed with the alternative ingress ip.


dguido:
Closing in favor of #14949 which uses the more correct approach: using the `is ansible.utils.ipv4_address` test (returns native boolean) rather than my filter + `| bool` workaround. Their `| trim` also handles potential whitespace from the metadata endpoint.

